PR #248, opened 2014-10-17
Text replacement rules

janza:
Added support for replacement rules on incoming messages.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -783,6 +783,10 @@ button {
 	display: block;
 	padding: 5px 0 10px 1px;
 }
+#settings .control-label {
+	display: block;
+	margin-top: 11px;
+}
 #settings .opt input {
 	float: left;
 	margin: 4px 10px 0 0;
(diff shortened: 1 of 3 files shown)


JocelynDelalande:
> I would find this useful (creating quickie shortcuts like /ns -> /msg nickserv, other shortcuts of stuff)

What you say is out of the scope of this PR (which deals with incoming messages).

JocelynDelalande:
I do agree with @erming it's not useful enough to get into core. @janza maybe in a plugin as suggested before, sorry that plugins are not available yet and there is no immediate option for your feature @janza thanks for submiting your code :)

Closing, @astorije feel free to re-open and comment if your opinion differs.

astorije:
I believe everyone agrees with each other here, even the OP :-)
In this thread, you read very specific use cases, that apply to very few people. That indeed makes us lean towards a plugin system. Let's talk about such plugin system in our [future roadmap](https://github.com/erming/shout-roadmap).

@janza, so sorry we have to close your PR, but that doesn't mean we do not welcome the idea overall (in a plugin, that is). Also, please be sure that we appreciate the effort and hope to see you back in the Shout community :-)

Issue #35, opened 2014-08-18
Pre-fetch links from chat messages


erming:
Inspired by how Facebook handles links.

When someone posts a link, I want to pre-fetch that link and display information directly in the chat.

This could then be extended to work with:
- Thumbnails
- YouTube movie information
- GitHub info
- ...

astorije:
[Telegram](https://telegram.org/) also does this wonderfully.
However, I am not too fond of using an external API for this...

@erming, this seems like it's done in current version (0.51.2), closing this issue to help with triage then. Feel free to re-open if I'm mistaken, but I feel like it's more relevant to open specific topic issues for this.

PR #275, opened 2014-11-02
Expand links with Embed.ly

gaving:
Added support for expanding links/images with [Embed.ly cards](http://embed.ly/cards).

#35 relates, effectively reverts #228 and removes the `prefetch`, `links` and `thumbnails` options (plus a bunch of dependencies).

In action:-

screen

Settings:-

settings


spencerthayer:
@erming have you found the time to merge this.

JocelynDelalande:
:-1: huge privacy concerns about that, it logs every url to an external, centralized and proprietary service.

floogulinc:
[embed.js](https://github.com/ritz078/embed.js) might be a great alternative to this.

YaManicKill:
I'm assuming this isn't going in because of being a 3rd party service. It might worth adding as a plugin if that ever gets done, but this could probably be closed for now?

astorije:
Yes, make this a third-party package is the right route, as people need to be aware of the privacy concerns before committing to using that.

Issue #54, opened 2014-09-08
Add shout to docker


furier:

mishak87:
Hi how is it going?
Could I be of any help? I got experience with docker, just based home server on docker-only setup and shout looks like superb IRC client to have there.

I would like to use official distribution packed by automated builds.

For docker image it needs more ENV parameters with reasonable defaults, but there sed with init script can come to rescue.

prologic:
Let's get this merged and published to the hub already!

flaccid:
+1

astorije:
This was added to Shout, closing this issue.

PR #285, opened 2014-11-10
Fixing issues #54: Add shout to docker

PierreZ:
Wrote thank's to @furier a valid Dockerfile valid for private and public shout server!

README.md in data/{logs,users} are just here to create the hierarchy.

All the commands are in the main README.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+# Based on the work of Furier, thank's to him! https://gist.github.com/furier/09f1bbe309e2c19236a1
+ 
+FROM node:latest
+ 
+# set user root
+# using root user to bind volume without permission's issues
+USER root
+ 
+# install shout
+RUN npm install -g shout
+ 
+# expose server port
+EXPOSE 9000
+ 
+# entrypoint is shout, to start the shout server...
+ENTRYPOINT shout
--- a/README.md
+++ b/README.md
@@ -18,6 +18,31 @@ __What features does it have?__
 sudo npm install -g shout
 ```
 
+## Install with Docker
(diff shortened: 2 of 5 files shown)


ryanseys:
The current Docker instructions [here](http://shout-irc.com/docs/deployment/docker.html) reference a Dockerfile that does not exist. This PR should fix that.

czak:
How about a `Dockerfile` which actually uses the repo code?

```
FROM node:latest

COPY . /src 
RUN npm install -g /src
 
EXPOSE 9000
 
CMD ["shout"]
```

Differences from the Dockerfile proposed in this PR:

* `USER root` is not needed, `root` is [default in the container](http://docs.docker.com/reference/run/#user)
* copy and install from the source in current folder (i.e. the repository)
* `CMD` instead of `ENTRYPOINT` so you can override the command in `docker run`

That's how I set it up in [my fork](https://github.com/czak/shout). Then you can set up an [automated build](docs.docker.com/docker-hub/builds/) on Docker Hub so everyone will be able to:

```
$ docker run -p 9000:9000 erming/shout
```

...without building the image themselves. And it will stay up to date on Docker Hub.

> I don't think you need that data folder. 
> You can just mount /root/.shout from container to any host directory and the application will create the needed structure.

:+1: to this. With docker everyone can place the configs wherever they like.

hashworks:
Just a tip: You can refere to issues by adding stuff like "resolves #54" to your pull request description. This way the issue will also get closed automatically when this get's merged. Magic!

PierreZ:
@czak is right. I'm closing my PR

PR #311, opened 2014-12-18
convert gif links to videos via Gfycat

vohof:


hashworks:
I would suggest `Convert .gif links to webm format` instead of `Convert .gif links to videos`.

astorije:
Hi @vohof, thanks for your contribution, we greatly appreciate the effort :-)

However, I must say, 500+ lines worth of dependency, that's heavy! May I ask why converting GIFs into videos instead of... keeping them in GIF?!
What are the benefits, the drawbacks, the consequences, ...

Your PR is really big but you haven't provided any information on it, we would happily discuss before considering a merge :-)

astorije:
@vohof, you closed your PR, but I'm still very interested to know why one would convert GIFs to videos in our case. Care to explain? :-)

Thanks again for taking the time to stop by here!

vohof:
@astorije Videos tend to load faster than GIFs. The big major file is taken from: https://github.com/gfycat/gfycat.js

astorije:
Hum, I have some doubts about our needs with this. This seems more like an optimization, and I'm not sure Shout reached that point yet. @erming, what's your take on this?

I would say that we reconsider this when major issues are taken care of, as this adds future maintenance.

However, this sort of PRs shows me that we need some smart thinking around client-side library management...

PR #320, opened 2014-12-28
Client plugins

janza:
Support for loading additional scripts on frontend.
Added basic hooks for plugins to client script.

Scripts can be added to ~/.shout/plugins directory.

Registering a plugin:

```
shoutPlugins.push({
    onmessage: function(message) { console.log(message); },
    onprerender: function(name, data) { console.log(data); },
    onppostrender: function(name, html) { console.log(html); }
})
```

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -417,6 +417,9 @@ $(function() {
 			clear();
 			return;
 		}
+		text = pluginHandle('oninput', text);
+		if (!text) return;
+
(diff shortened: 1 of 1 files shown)


YaManicKill:
@astorije Would this work as a base for the package system? It might have quite a few merge conflicts because it has been over a year, but might be worth starting if @janza is worth updating it? It looks like this was never looked at.

PR #329, opened 2015-01-03
Catch undefined content-type headers

linux4kix:
If content-type is undefined then catch the exception and set it
to empty just as is done for the body.


erming:
Hey! Thanks!
I've pushed a new version (`0.49.1`) with this fix.

PR #332, opened 2015-01-07
Pushbullet Integration

japesinator:
This adds pushbullet integration for notifications.  At the moment, config is in the global config, as I prefer that for sensitive tokens, but I would be open to moving it into user config if that is preferred.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -241,7 +241,7 @@ module.exports = {
   // Run Shout with Pushbullet support
   //
   // @type     object
-  // @default  false
+  // @default  {}
   //
   pushbullet: {
 
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -236,4 +236,29 @@ module.exports = {
 		//
 		port: 113
 	}
+
+  //
+  // Run Shout with Pushbullet support
+  //
+  // @type     object
+  // @default  false
+  //
+  pushbullet: {
+
+    //
+    // Run Shout with Pushbullet support
+    //
+    // @type     boolean
+    // @default  false
+    //
+    enable: false,
+
+    //
+    // User/Pushbullet access token pairs
+    //
+    // @type     object
+    // @default  {}
+    //
+    tokens: {}
+  }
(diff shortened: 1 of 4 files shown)


astorije:
Hi @japesinator, thank you very much for your contribution!

This seems to me however pretty bad for privacy, as it relies on third-party that can record and re-use your data, even in private messaging. Also, using OTR to encrypt the messages is not possible as this Pushbullet integration (and I'm guessing Pushbullet altogether) relies on unencrypted messages.
@JocelynDelalande usually has a broader view of privacy consequences, I would like to hear his opinion before considering more work on this or closing the PR.

By more work, I mean that we cannot expect administrators to manage the user tokens. But please do not waste your time on improving this, it would be a shame if we decide in the end to refuse that PR, which is likely.

I know I have been saying this a lot lately, but I am more inclined to consider this as a server-side plugin. It would be the decision of the network administrator to enable this plugin and communicate with his users about the potential dangers of this.
In the meantime, since the plugin system is under discussions, I invite you to join us on https://github.com/erming/shout-roadmap if you have opinions on doing so.

floogulinc:
Users should have to use oauth to sign into their pushbullet account and then we store the token in the user config file. We also have to have the app token in the global config. See https://docs.pushbullet.com/#oauth2

Also, we may be able to use Pushbullet's End-to-End Encryption, but I haven't looked much into it.

JocelynDelalande:
:-1: for putting this feature into core project, well suited for an external plugin.

@astorije agreed on privacy concerns with 3rd party service (still ok as a plugin, people do what they want), 

@floogulinc interesting pointers even if I didn't took time to realy read, @japesinator do you know those APIs/technologies ?

@japesinator thanks for your proposal anyway, the thing is: we cannot integrate it to core, and plugin system is not functional at the moment, so sorry but it's a no-go for now :-/. Feel free to contribute on plugin needs as @astorije suggested.

(closing according to @astorije @floogulinc and my opinion)

PR #337, opened 2015-01-13
Added FastClick to eliminate the 300ms delay between a physical tap and ...

tditlu:
...the firing of a click event on mobile browsers


vohof:
:+1:

vohof:
I don't seem to be seeing the effect (ios7), still has the delay.

tditlu:
I'm on 8.1.2, but a good e.g. is to try opening the off-canvas menu and try to click on the "Connect to network" + button. Without FastClick this require 2 clicks/touches...

erming:
Hey, sorry for slow answer!

However, during this time, I've found another solution to this problem:

```
$("button").on("touchstart click", function(e) {
  e.preventDefault();
  // Do your things here.
});
```

It listens to both `click` and `touchstart`, but the `e.preventDefault()` makes sure that `click` doesn't trigger after `touchstart` was triggered.

This tiny solution might be a better fit than loading another external library. Not sure about different browser support, but it works on iOS.

tditlu:
If it works on Android it's a much cleaner solution than mine!

astorije:
Does someone wants to confirm @erming's solution and open a PR for this?

@erming, @tditlu, considering that alternative solution, should we close that current PR?

astorije:
I am also reading [this](https://github.com/ftlabs/fastclick#when-it-isnt-needed) which essentially says this is not needed for Chrome on Android.

Finally, if we still decide to go for something like this, then we should be using a package manager instead of including the lib directly.

astorije:
I am contemplating closing this as it might not be needed anymore and no one has spoken up about my last comment nor complained on IRC/issues.

@tditlu, @erming, are you OK with this?

tditlu:
Ok for me!

astorije:
@tditlu, do you confirm this is not needed anymore? :-)

YaManicKill:
Most touch browsers now hide the delay when you add 

`<meta name="viewport" content="width=device-width">`

Which we already have. There's probably a few edge cases but it feels unnecessary if no one has brought that up recently. 

I certainly don't feel the delay on chrome on android.

astorije:
Okay, closing then!

PR #344, opened 2015-01-22
Prevent HTML injection through /topic!!!!

PangeaCake:
Really big security issue here.


PangeaCake:
untested though, beware

erming:
Oh, already fixed this issue (here https://github.com/erming/shout/commit/ef041d6612e20369bf49b17b8162c57b4955a1cc) before I saw this. But I might `_.escape` topic anyway. Merging.

PR #348, opened 2015-01-27
Use document title to notify users of new messages

jriddick:
Hello,

This pull-request implements a feature from irccloud that makes it easier to see if you have gotten any new messages without focusing on the window. It adds a simple plus sign (+) in the title whenever you get a new message and when you give the window focus it will be removed.


client/index.html
```diff
@@ -286,6 +286,13 @@ <h1 class="title">Settings</h1>
 	<script src="js/libs.min.js"></script>
 	<script src="js/shout.templates.js"></script>
 	<script src="js/shout.js"></script>
-
+    <script src="js/ifvisible.js"></script>
+    <script>
+        ifvisible.on("focus", function() {
```
YaManicKill: No real need for this to be in a script tag here, is there? Couldn't it be in js/shout.js.

client/index.html
```diff
@@ -286,6 +286,13 @@ <h1 class="title">Settings</h1>
 	<script src="js/libs.min.js"></script>
 	<script src="js/shout.templates.js"></script>
 	<script src="js/shout.js"></script>
-
+    <script src="js/ifvisible.js"></script>
```
YaManicKill: This should be in js/libs/ and will be included in libs.min.js

client/js/ifvisible.map
```diff
@@ -0,0 +1,19 @@
+{
```
YaManicKill: What is this file for?

YaManicKill:
1 year after opening...

I like this feature a lot, but obviously some things would have to change as things have changed quite a bit in the last year (mainly the formatting, and the library)

astorije:
I agree with @YaManicKill's comment. The proposed feature is great but needs mentioned improvement.
@jriddick, would you be willing to give another look at this please? :-)

I'm not too fond of having a dependency telling us if the tab is currently opened. I believe modern browsers (which is what we are targeting anyway) are pretty reliable in that regard.

YaManicKill:
Just fyi, http://www.thefutureoftheweb.com/blog/detect-browser-window-focus

astorije:
Exactly, I don't think there is a need for a library for this.
Also https://developer.mozilla.org/en-US/docs/Web/API/Document/hasFocus if I'm not being mistaken.

PR #357, opened 2015-02-06
use users icon for user list instead of second menu icon

jancborchardt:
Before, on mobile there were 2 menu icons, one on the left and one on the right. This makes it clearer what’s behind the right icon.

The icon doesn’t look particularly awesome but that’s Fontawesome …

Please review @erming

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -122,6 +122,9 @@ button {
 	font: 14px FontAwesome;
 	content: "\f0c9";
 }
+#viewport .rt:before {
+	content: "\f0c0";
+}
 #viewport .rt {
 	display: block;
 	float: right;


dideler:
Hey @jancborchardt, thanks for the PR!

Please consider adding before and after screenshots for visual changes. In some cases a gif works well.

The easier it is to review a patch, the more likely it is to get merged :)

jancborchardt:
Sure thing! Before:
shout-before
After:
shout-after
(As said above the icon looks sub-par but that’s due to the icon set used.)

erming:
Hey!
Thanks for showing interest in helping out.

I initially used that icon, but I personally don't like the looks of it. I experimented with other some icons, but I liked the hamburger too much. ;)

So I am sorry: I will not merge this.

However, if you find another icon that looks sweet, feel free to suggest it!

Thanks!

PR #367, opened 2015-02-26
Allows to get default user data from HTTP Headers

lucamilanesio:
When using shout behind a reverse-proxy with possibly user-authentication
(scenario of using it within an enterprise auth-proxy) allows to get
the default pre-filled user identity from HTTP headers.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -70,6 +70,15 @@ module.exports = function(options) {
 	}
 };
 
+function reqHeader(req, header, defValue) {
+	var headerValue = req.get(header)
+	if (headerValue) {
+		return headerValue
+	} else {
+		return defValue
+	}
+}
+
 function index(req, res, next) {
 	if (req.url.split("?")[0] != "/") return next();
 	return fs.readFile("client/index.html", "utf-8", function(err, file) {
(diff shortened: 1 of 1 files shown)


astorije:
@JocelynDelalande, you have a much deeper understanding of networks and HTTP protocol than me, can you give your thought (and any other brain dump that might be useful to record?) on this?

Also, does that look like a bug fix or an enhancement to you?

JocelynDelalande:
@lucamilanesio that's a good feature, definitely useful when using some SSO. Thanks :)

Next step (I mean, in another PR) would be to do (optional and disabled by default) external auth : user login is provided in HTTP headers by reverse-proxy. I opened #482 just to keep it in mind.

@lucamilanesio  two questions:

- could you think about somewhere to document that feature ?
- out of curiosity, how did you choose the headers names ? Is there any de-facto standard ?

src/server.js
```diff
@@ -70,6 +70,15 @@ module.exports = function(options) {
...
 
+function reqHeader(req, header, defValue) {
+	var headerValue = req.get(header)
+	if (headerValue) {
+		return headerValue
+	} else {
+		return defValue
+	}
+}
+
```
JocelynDelalande: I think that helper is unnecessary, you can handle the whole thing with a 3rd dict argument to [this `_.merge()`](https://github.com/lucamilanesio/shout/blob/get-default-user-identity-from-header/src/server.js#L85)

JocelynDelalande:
@lucamilanesio bump ?

xPaw:
Why not simply use HTTP authentication? I don't see the need to implement custom headers.

JocelynDelalande:
> Why not simply use HTTP authentication? I don't see the need to implement custom headers.

@xPaw It's not the same. Here we're getting default user information by a third-party (generally, the reverse-proxy itself, which may use HTTP basic/digest), then that third-party populate the custom HTTP header with informations about the user it knows.

What I suggest for longer-term and another PR (external auth) is the same thing except we're not only pre-populating form data but actually logging-in the user by this mean.


@xPaw I don't know if I'm clear in my explanations… am I ? :)

PR #343, opened 2015-01-21
WEBIRC support.

bopjesvla:
WEBIRC support.

Issue #318, opened 2014-12-25
Use client connection, not the server's


rollecode:
Hi,

I'm using mainly Quakenet which has connection limits (5 per ip address and 2 per username). It's possible to have Quakenet trusts with specific IP, but best practice would be to have the app use client's connection and hostname, not the server's. Qwebirc has this feature, but it's not nearly as fancy, not to mention reponsiveness and other goodies shout has.

```
Too many connections
```

Any ideas?

Thanks already.

Roni

spencerthayer:
+1

heriniaina:
+1

Jnbond:
+1

vlanmx:
Guys, any update on this?

Lack of webirc support makes this project very unuseful.

PR #379, opened 2015-03-21
implemented WEBIRC

AmShaegar13:
I've seen #343 which adds WEBIRC support as well but it does not do it right. Their method will always use the IP address of the first or last client connected to shout, can't remember which one.

My method is based on the client's real socket and thus will use the correct ip.

I felt like sharing it might help you.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -92,17 +92,17 @@ function init(socket, client, token) {
 		socket.emit("auth");
 		socket.on("auth", auth);
 	} else {
-		var ip = socket.handshake.headers['x-forwarded-for'] || socket.request.connection.remoteAddress;
+		var ip = config.reverse_proxy ? socket.handshake.headers['x-forwarded-for'] : socket.request.connection.remoteAddress;
 		client.host.ip = ip;
 		try {
 			dns.reverse(ip, function(err, clienthost) {
-				if(!err && clienthost.length) {
+				if (!err && clienthost.length) {
 					client.host.hostname = clienthost[0];
 				} else {
 					client.host.hostname = ip;
 				}
 			});
-		} catch(e) {
+		} catch (e) {
 			client.host.hostname = ip;
 		}
 
--- a/src/server.js
+++ b/src/server.js
@@ -5,6 +5,7 @@ var ClientManager = require("./clientManager");
 var express = require("express");
 var fs = require("fs");
 var io = require("socket.io");
+var dns = require("dns");
 var Helper = require("./helper");
 var config = {};
 
@@ -91,6 +92,19 @@ function init(socket, client, token) {
 		socket.emit("auth");
 		socket.on("auth", auth);
 	} else {
+		var ip = socket.handshake.headers['x-forwarded-for'] || socket.request.connection.remoteAddress;
+		dns.reverse(ip, function(err, clienthost) {
+			var hostname;
+			if(err || !clienthost.length) {
+				hostname = ip;
+			} else {
+				hostname = clienthost[0];
+			}
+		
+			client.host.hostname = hostname;
+			client.host.ip = ip;
+		});
+
 		socket.on(
 			"input",
 			function(data) {
--- a/src/server.js
+++ b/src/server.js
@@ -93,17 +93,18 @@ function init(socket, client, token) {
 		socket.on("auth", auth);
 	} else {
 		var ip = socket.handshake.headers['x-forwarded-for'] || socket.request.connection.remoteAddress;
-		dns.reverse(ip, function(err, clienthost) {
-			var hostname;
-			if(err || !clienthost.length) {
-				hostname = ip;
-			} else {
-				hostname = clienthost[0];
-			}
-		
-			client.host.hostname = hostname;
-			client.host.ip = ip;
-		});
+		client.host.ip = ip;
+		try {
+			dns.reverse(ip, function(err, clienthost) {
+				if(!err && clienthost.length) {
+					client.host.hostname = clienthost[0];
+				} else {
+					client.host.hostname = ip;
+				}
+			});
+		} catch(e) {
+			client.host.hostname = ip;
+		}
 
 		socket.on(
 			"input",
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -256,5 +256,17 @@ module.exports = {
 	//
 	webirc: {
 		"irc.example.org": "yourwebircpassword"
-	}
+	},
+
+	//
+	// If you use WEBIRC:
+	//
+	// Set this option to true if you run Shout behind a reverse proxy. Do not
+	// enable this option if Shout is accessed directly. This would allow users
+	// to fake their IP address which makes them immune to bans.
+	//
+	// @type     boolean
+	// @default  false
+	//
+	reverse_proxy: false
 };
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -246,5 +246,15 @@ module.exports = {
 		// @default  113
 		//
 		port: 113
+	},
+
+	//
+	// Send WEBIRC password for defined host.
+	//
+	// @type     object
+	// @default  {}
+	//
+	webirc: {
+		"irc.example.org": "yourwebircpassword"
 	}
 };
--- a/src/client.js
+++ b/src/client.js
@@ -74,6 +74,8 @@ function Client(sockets, name, config) {
 	}
 }
 
+Client.prototype.host = {};
+
 Client.prototype.emit = function(event, data) {
 	if (this.sockets !== null) {
 		this.sockets.in(this.id).emit(event, data);
@@ -157,6 +159,10 @@ Client.prototype.connect = function(args) {
 
 	var irc = slate(stream);
 	identd.hook(stream, username);
+
+	if (config.webirc[server.host]) {
+		irc.write("WEBIRC " + config.webirc[server.host] + " " + username + " " + this.host.hostname + " " + this.host.ip);
+	}
 
 	if (args.password) {
 		irc.pass(args.password);


ronilaukkarinen:
Tried it, but I guess it won't work with Quakenet. Been trying all of these but no luck with Quakenet. Too bad, because this looks too good compared to webchat.quakenet.org. Note #318.

AmShaegar13:
What did you try? Do you have a WEBIRC password for Quakenet?

ronilaukkarinen:
To be honest, I'm not sure how to use WEBIRC with Quakenet, just left the password empty. That's why I'm asking if someone would know or point me to the right direction. I think I'm going to open a new issue about Quakenet, since I'm not sure if it supports WEBIRC after all.

AmShaegar13:
Giving you a WEBIRC password means that they fully trust your word. You can freely set your users host name which is the essential for banning them.

ronilaukkarinen:
Ok, maybe the WEBIRC is not the solution then since Quakenet is a public server, but I'm wondering how [qwebirc](https://qwebirc.org) has managed to achieve that. Anyway, the ident and hostname are staying the same regardless the shout username - which makes Quakenet to make a G-line max users ban when 2 or more users are connecting through shout...

AmShaegar13:
> qwebirc is [an] IRC client designed by [...] the QuakeNet IRC network. 

I guess they either have a WEBIRC password or have disabled connection limits for this special host.

spencerthayer:
Does WEBIRC work? Can we get it merged?

spencerthayer:
+1!

AmShaegar13:
After running this for about a month now I encountered an error which crashed Shout completely. http://pastebin.com/ENRh9yLK

Looks like we need a try/catch there around the reverse() method.

spencerthayer:
Damn. Well I suppose this isn't ever going to get merged with the project.

spencerthayer:
@AmShaegar13 did you resolve you error? Is Shout now perm broken on your end?

spencerthayer:
Hey @erming can you please merge this implementation of WEBIRC (unless you've integrated a better solution) to the master branch?

AmShaegar13:
Hey,
I did not surround this by try/catch yet to see if it crashes again. It didn't. But for a production software I'd include it anyway.

Let me fix that. You could merge that manually if you need it.

AmShaegar13:
Added the try/catch and rebased the branch on the current master.

spencerthayer:
Great. As soon as @erming gets this in the master we can start using Shout on hosted servers without violating most IRC servers TOS! Which means I can use Shout again in general.

tehsu:
If shout is restarted and a user reconnects they end up getting the Server IP instead of the Client IP.

JocelynDelalande:
@AmShaegar13 Can't find what is this webirc thing, any doc/pointer on that ?

AmShaegar13:
This article should explain it very well: https://www.unrealircd.org/docs/WebIRC_Support

Because shout lacks support for this, it is mostly useless when used for communities larger than 5 people on a public IRC network.

Xe:
This also requires setup on the irc server you are connecting to. I don't suggest merging this pull request as it will cause confusion.

JocelynDelalande:
@AmShaegar13 ok, thanks :)
@Xe yes, but it seems lgtm feature to me, no ?

A privacy issue is that your IP will unexpectedly appear on IRC server where the user may expect to be proxied behind shout.

Xe:
No, it won't work 100% of the time. You have to explicitly set this up with the target IRC server. It is not just a magic "this will work automatically".

JocelynDelalande:
> No, it won't work 100% of the time. You have to explicitly set this up with the target IRC server. It is not just a magic "this will work automatically".

Yes, I already understood that.

@AmShaegar13 your first comment seems to say that it's difficult to use because the big IRC networks limiting the number of users equally refuse to give webirc passwords, in this situation, does is actually unlocked some situations for you ?

AmShaegar13:
We planned to create a community for an online game and asked some IRC networks if they would grant us WebIRC access. We were given an individual password to run shout on one of those networks.

That's the reason why we added support to shout and when I saw there was demand for that feature I decided to share our implementation. 

So yes, WebIRC helped us start the project.

know75:
Speaking as the owner of a medium sized IRC network (there are a lot of us out there), and as a great fan of Shout I believe it's essential that WEBIRC at least be a configurable option within its core. If people don't want to use it, they don't have to. 

As Shout is a multi user resource, without WEBIRC it makes it impossible to ban (either from a channel or a whole network) any individual. It's all or nothing. 

IRCCLOUD doesn't use WEBIRC, but creates an unique ident/username for each user; unfortunately this is open to abuse as people constantly ban evade by signing up for a new account and thus get a new ident/username. I'd much rather add a simple WEBIRC setup to my IRC servers' config files than have to ban all users. 

In the case of IRCCLOUD we've had to ban the whole client. I'd hate for the same to happen to Shout. We have WEBIRC set up for both Kiwiirc and Mibbit and don't experience any problems.

Xe:
You raise a point. I think there is a better way this could be implemented though.

AmShaegar13:
Name it please. I don't see it and so do network owners. Being able to ban individuals based on their IP seems to be essential. The way to go to pass that IP through the client to the server is WebIRC.

tehsu:
Network owner here, please merge this.

Xe:
It should at least have the IP and reverse DNS it spoofs be configurable per user.

AmShaegar13:
I am not sure if I get you right. Do you want the Shout user configure its own IP? That would make this feature rather useless. You still can't identify individuals and issue IP bans then.

Or is it privacy you are concerned about? Should Shout mask the users data irreversible based on their IP? I think this is rather a network than a client feature.

kalaksi:
@know75 , how does WebIRC work any better than the solution used in IRCcloud? If the problem is that people create new fake accounts, what prevents them from creating new shout accounts, too?

Users can easily obtain new IP addresses or it may change automatically (dynamic, mobile, VPN etc.), so IMO the best bet would still be to look at the username (assuming ident is used) and contact the provider about the abusive user.

kalaksi:
@AmShaegar13 if shout doesn't store the last IP address of the user for WebIRC, what address does it use when user isn't connected to shout, but shout loses connection to the IRC server and reconnects?

AmShaegar13:
I am really sorry, but I don't understand enough of how Shout works with registered users as I never used it myself. :confused:

kalaksi:
@AmShaegar13 ah, I misread the first comment, you meant the first or last client, not the first or last address of a client. Anyway, it's still relevant (for registered users) as shout keeps the connections to IRC constantly open. I haven't used shout without registered users so I didn't know it also works without having persistent connections.

Xe:
@AmShaegar13 I mean the administrator being able to add configuration to a user such as:

```json
"webirc": {
    "ip": "25.0.0.1",
    "rdns": "bouncer/user/foobar"
}
```

know75:
First of all, there is no ideal way to ban someone. With every system there are ways to overcome the method employed. The best we can do is allow ourselves open to as many options as possible. In this case, as I said in my original post, having the option of being able to use WEBIRC affords us another way in which to ban someone. WEBIRC should be an option available to the Shout owner; no one is forcing them to use it but from the all the posts here it seems a great many people would like WEBIRC.

@kalaksi: 

> how does WebIRC work any better than the solution used in IRCcloud? If the problem is that people create new fake accounts, what prevents > them from creating new shout accounts, too?

Because their IP/hostname won't change, only their ident/username.

> Users can easily obtain new IP addresses or it may change automatically (dynamic, mobile, VPN etc.), so IMO the best bet would still be to > look at the username (assuming ident is used) and contact the provider about the abusive user.

Absolutely, there are many ways for someone to change their IP. However, most networks have some defense against such tactics. For example, on connection IRCds/Services can check for TOR users, open proxies and ofter check the various IP blacklists that are freely available and kept up to date. As for dynamic hostnames, wildcards are used, for example *!*@123.456.* Without WEBIRC you cannot ban dynamic IPs. 

I would suggest that we use both the ident/username and WEBIRC (its usage to be determined by the Shout owner) methods to offer as many ways as possible to combat trolls and whatnot. 

You also mention that the Shout owner should delete users that abuse the system. I encourage that too, but when you're in the midst of a troll attack from 4chan you then have to reach out to the Shout owner to delete the 1000s of fake accounts/connections being created -- you're adding another step and delay in resolving the problem. Channel Operators (with WEBIRC enabled) should be able to take all the steps they can to control the situation. Without WEBIRC, they're dealing with just one IP and would have to ban all users using Shout.

@kalaksi:  (second post)

> Edit: @know75 if you were also talking about a case with anonymous users, many webirc-clients I have seen generate an id from user's IP > address to use it as a username. That wouldn't require hostname spoofing or passwords.

This is called cloaking. https://www.unrealircd.org/docs/Cloaking The real IP of the user is passed onto the IRCd via WEBIRC and then is cloaked via the IRCd. Cloaking can't work without WEBIRC, or to be precise it would work, but everyone would have the same cloaked IP (that of the server Shout is based on).

@Xe If I understand you correctly, you're suggesting Shout assigns an IP/rDNS to each user. I see two problems with this:

1) The same with ident/username banning, when a new fake account is created then a new IP/hostname will be assigned.
2) As mentioned by AmShaegar13, WEBIRC permissions are granted by network owners on trust. I honestly doubt many networks would agree to a WEBIRC agreement if they know you're going to be sending incorrect information to them via this protocol. 

To summarize, I do believe in ident/username banning, but I also think it essential that WEBIRC is an option that Shout owners can put into practice. I'd hate for this project to reach completion only to find that networks won't allow its use.

JocelynDelalande:
> Users can easily obtain new IP addresses or it may change automatically (dynamic, mobile, VPN etc.), so IMO the best bet would still be to look at the username (assuming ident is used) and contact the provider about the abusive user.

Well, that's not *that* easy to get *a lot* of IP addresses.

The feature seems legitimate to me. I just have a concern:

> Or is it privacy you are concerned about? Should Shout mask the users data irreversible based on their IP? I think this is rather a network than a client feature.

IMHO, it's what you expect when behind a tool like shout (to be hidden), and it's what happens *most of the time*. 

@AmShaegar13 (and others), what do you think about a way to present a warning/info to the user when before she/he joins a network with WebIRC enabled that his/her real IP will be revealed to the world ?

AmShaegar13:
Personally, I was rather surprised that Shout did not pass my own IP to the network when I first used it. I would have expected it do so. But that's due to the fact that I am familiar with banning, trolls etc.

I wont deter you from adding such a warning to WEBIRC-enabled instances of shout, although I expect users to raise privacy concerns if they get such a warning which they would never think of without that warning. That's how IRC works and always worked. Desktop clients do not show a warning to users; why should Shout?

JocelynDelalande:
> Personally, I was rather surprised that Shout did not pass my own IP to the network when I first used it. I would have expected it do so. But that's due to the fact that I am familiar with banning, trolls etc.

Yeah, and I'm more used to peaceful chans, so not familiar with those issues, that's why your opinion is valuable :)

> I wont deter you from adding such a warning to WEBIRC-enabled instances of shout, although I expect users to raise privacy concerns if they get such a warning which they would never think of without that warning.

Yup, but on the spirit, I prefer people to be aware and yelling than ignorant (eventually, it will yltimately guide them towards real protection mechanisms).

> That's how IRC works and always worked. Desktop clients do not show a warning to users; why should Shout?

Because it's not a desktop app but a web client, so you are connecting "behind" this client.

Anyway… That's my point of screen+irssi user ; but looking at other web clients, you seem to be right, that's the way it's done (showing the browser client IP address).

So… I don't know what to think about it actually. Maybe if I'm the only one raising this as a privacy issue, merging as-is is ok for now.

kalaksi:
@know75 I see your point. I believe having different choices is good and I think this would be useful. 

I'm more used to seeing user's IP set to username (hex-encoded) in webirc-clients. I haven't used that many, though (and frankly, why not make the whole client run on the browser instead?). It could be used for range-banning and be simpler as it wouldn't require WEBIRC or requesting passwords and could therefore be more useful on public servers. But where available, WEBIRC would probably be a better solution. I'm just thinking out loud.

@JocelynDelalande I consider the functionality for registered users similar to a bouncer or screen+irssi (as you probably read from the other PR) and therefore I wouldn't assume it to forward the user's IP address by default. As I see it, regular webirc-clients always need to have WEBIRC-functionality (or something similar) for battling abuse because they don't require user registration and the connection to IRC isn't usually initiated from the user's computer. This applies to anonymous use of shout, too, apparently (haven't used myself). 

Anyway, I think this PR should take registered users and persistent connections into account in some way, too.
Regarding the warning, I think a simple note would be enough instead.

src/server.js
```diff
@@ -91,6 +92,20 @@ function init(socket, client, token) {
 		socket.emit("auth");
 		socket.on("auth", auth);
 	} else {
+		var ip = socket.handshake.headers['x-forwarded-for'] || socket.request.connection.remoteAddress;
```
JocelynDelalande: I get the impression, reading [shout documentation](http://shout-irc.com/docs/) and the way it is designed that deployment without reverse-proxy is one of the options offered, in this case, `x-forwarded-for` can be spoofed by browser ; dare to add in this PR, an option to settings to declare that shout is being ran behind a reverse proxy ?

Xe: If you are exposing node.js to the world on port 80 without a proxy you are pretty much inviting what happens.

AmShaegar13: Yep. You are right... Looks like you need a list of trusted proxies. Or a simple toggle in configuration if you are behind a proxy or not.

I will remove that from this PR and maybe create another one to add that feature properly implemented again.

JocelynDelalande: > Yep. You are right... Looks like you need a list of trusted proxies. Or a simple toggle in configuration if you are behind a proxy or not.

The second solution (the toggle in configuration) seems simple and fine to me.

> I will remove that from this PR and maybe create another one to add that feature properly implemented again.

IMHO, better to stick on this very PR, reworking it. It's easier to continue conversation, and avoid polluting the bug tracker.

AmShaegar13: Ok.

apmorton:
I believe quite a few people in this thread are misunderstanding the use of WEBIRC - this is not a feature someone would enable on their own if they were running shout for personal use.

This is a feature that would exclusively be used by an administrator of an IRC network who was trying to setup a public web based irc client for their own network.

take the freenode webchat for example. https://webchat.freenode.net/

The following is what shows up when you connect to freenode through their web client.
```
nick [ident@gateway/web/freenode/ip.{client ip here}]
```

I would say personally that this feature should be merged as is with one change.  Entirely disable it when `public` is disabled.  I see no value in using this feature in combination with persistent user accounts - as in my mind they are mutually exclusive.

You are either setting up shout for personal use, or you are setting it up for public use and you REQUIRE webirc (or another solution such as cgiirc) to identify real users on the other end of shout.

If you are using a publicly hosted instance of shout there is absolutely no expectation of privacy.

JocelynDelalande:
@apmorton thanks for writing things down, it sometimes help  :)

> I would say personally that this feature should be merged as is with one change. Entirely disable it when public is disabled. I see no value in using this feature in combination with persistent user accounts - as in my mind they are mutually exclusive.

Makes sense but…

> You are either setting up shout for personal use, or you are setting it up for public use and you REQUIRE webirc (or another solution such as cgiirc) to identify real users on the other end of shout.

There is a third use-case you forget: you host shout for a small organization/group of people, and provide accounts to your users. That's where WEBIRC is legitimate in private mode. 

Here WEBIRC would trigger an unexpected exposure of your IP address, so I would say for now, let's disable webirc for public mode (shout should fail on startup if public mode + webirc are enabled), as @apmorton suggested and allow ourselves to reconsider if we see incoming users feedback asking to allow it.

@AmShaegar13 Does that proposal, does it fit your need ? What do you think ?

apmorton:
> There is a third use-case you forget: you host shout for a small organization/group of people, and provide accounts to your users. That's where WEBIRC is legitimate in private mode.

I disagree with this.  If you are hosting for a small group of people you likely expect the behavior is that IRC shows the IP of shout - and not your own.  In fact, webirc will not work at all unless you have authorization from the irc network you are connecting to - which is very rare given this basically allows you to tell the IRC network anything you want about where you are connecting from.

WEBIRC is not meant for the purpose of 'spoofing' your connecting address to HIDE yourself - in fact it is quite the opposite.  It is meant as an extension of trust from the IRC server to a web irc gateway - in which the IRC admins have allowed the web gateway to pass along the original source IP of the user connecting.

JocelynDelalande:
> I disagree with this. If you are hosting for a small group of people you likely expect the behavior is that IRC shows the IP of shout - and not your own. In fact, webirc will not work at all unless you have authorization from the irc network you are connecting to - which is very rare given this basically allows you to tell the IRC network anything you want about where you are connecting from.

I agree with you (see my suggestion at the end of the comment you quoted). Legitimate here meant "may be your only way to connect some servers with that kind of setup/users number".

> WEBIRC is not meant for the purpose of 'spoofing' your connecting address to HIDE yourself - in fact it is quite the opposite. It is meant as an extension of trust from the IRC server to a web irc gateway - in which the IRC admins have allowed the web gateway to pass along the original source IP of the user connecting.

Perfectly ok with that, but the fact you re-explain it makes me think we misunderstood each other on something ;-)

know75:
I still believe that WEBIRC -- with both persistent and non persistent connections -- should be open and left for Shout admins to decide (why wouldn't you credit the Shout administer/s to be able to decide what is right for themselves?). WEBIRC is very commonly used, just look at Mibbit and KiwiIRC..

There is a problem of course with persistent connections + WEBIRC. If the connection between Shout and the IRCd is dropped, Shout doesn't remember the last IP of each user and can't reconnect. My suggestion for both persistent and non persistent connections is there is not an 'auto rejoin' because:

1) in this way persistent connections won't be without an IP. The user will have to manually rejoin thus providing their IP. This really shouldn't be too much of an inconvenience as connections shouldn't be dropped often given most server configurations these days.

2) if you have say 500 users when Shout goes down you're going to cause a huge load jump for both the Shout server and the IRCd to cope with. 500 users rejoin all at once will not only tax Shout and the IRCd but also trigger flood protection where enabled.

If persistent connections aren't going to have WEBIRC then they need an unique username to be able to ban certain users from channels.

spencerthayer:
Just to clarify WebIRC still isn't being implemented? As big a fan of Shout as I am I cannot use it for any application since not using WebIRC is a violation of the terms of agreement for every IRC network I would use this for. It's not really an issue that should be up for debate, WebIRC is an essential part of a robust IRC application.

apmorton:
Spencer,

I have been using shout as a public web chat interface for a project for
the last month or two. I had to hack in webirc support, among other things.

Shout is a great project, but it has several shortcomings that make it not
suitable as is for large scale public use. Webirc support is one of these
issues, but more importantly, it leaks memory like nobodies business.

I have spent the last few days tracking down sources of leaks, and after
reviewing the code I came to the conclusion that shout really just isn't
designed for public use, where hundreds of people may connect to a single
instance at once, and thousands over time.

As a result, I started removing code and have ended up with a version of
shout that entirely removes the 'private' functionality, and hopefully all
the leaks that cause constant crashes when shout is under load.

If there is interest, I can clean up the changes and release them as a fork.

Austin

know75:
I'd be extremely interested in such a fork, Austin.

floogulinc:
@apmorton making Shout work well in large-scale public instances is part of what we have been focusing on in [Shuo](https://github.com/Shuo-IRC/Shuo). I would be interested to see the changes in your fork.

apmorton:
@floogulinc off the top of my head, the major changes involved the following:

- pidfile support using npid, so at the very least crashed instances can be restarted easily
- removing 'private' config option and all code paths that only made sense with it enabled
- removing message storage entirely (where shout has infinite in-memory storage), although with a quick look it seems that shuo solved this by setting a limit in memory
- removing the concept of 'networks' - one client == one irc connection

essentially I tried to strip out anything that was not directly required when hosting a public instance

Xe:
@floogulinc you're getting his goal backwards. @apmorton wants KiwiIRC and doesn't know it. Shuo's idea was more for IRCCloud.

apmorton:
You are semi right about what i wanted.

Shouts UI, with kiwi's use-case.

Essentially I really liked shouts design, so I have adapted it's back-end
to suit my needs.

Xe:
Fair enough! Was worth throwing out just in case.

floogulinc:
@Xe Oh, crap, I was reading too fast and saw 
> removes the 'private' functionality"

as 
> removes the 'public' functionality"

spencerthayer:
@apmorton I would be very grateful for a public only WebIRC compatible fork of Shout!

spencerthayer:
@floogulinc does Shou work with WebIRC compliant servers? Also when you say large-scale public instances what does that mean? Does it remove the public features?

floogulinc:
@spencerthayer By "large-scale public instances" I am referring to running Shuo as a BNC service for an entire IRC network. Right now, this is pretty hard to do because it eats memory and loses connection to the IRC server without any notice and without reconnecting. WebIRC is not something we have implemented in Shuo yet.

JocelynDelalande:
> pidfile support using npid, so at the very least crashed instances can be restarted easily

@apmorton  Mind submitting this one as pull-request ?

spencerthayer:
@apmorton any chance we can see that WebIRC compatible fork of Shout? I could really use it right now. : )

itsrachelfish:
Why has this just been sitting open for over a year? It's the most requested feature by far and yet it seems like the shout developers don't care to implement it?

@erming ?

MaxLeiter:
@itsrachelfish The Lounge, the official fork of Shout, has had this implemented since May - https://github.com/thelounge/lounge/pull/240

astorije:
> Why has this just been sitting open for over a year? It's the most requested feature by far and yet it seems like the shout developers don't care to implement it.

Isn't open source a magical world? Countless of hours spent for free, and once in a while that friendly tone in return.
Note that WEBIRC has not been the most requested feature, not even remotely, neither on Shout or The Lounge. And also that activity on Shout has stopped some time ago and most developers have moved to The Lounge.

Anyway, as @MaxLeiter, The Lounge, as of v2.0.0+ (released a month ago) provides WEBIRC.

itsrachelfish:
@MaxLeiter @astorije Thanks for the tip! I'll check out lounge instead. :smile:

AmShaegar13:
Good to know. As Lounge's PR addresses some issues mine doesn't I'll close this one.

vlanmx:
Guys, any update on this?

Lack of webirc support makes this project very unuseful.

dgw:
WEBIRC support was merged into The Lounge some time ago. I won't reply to all of your copy-and-pasted comments about webirc in the Shout repo, but I highly recommend you look at switching. Shout is no longer developed, but The Lounge is an actively maintained fork.

vlanmx:
Thanks

PR #380, opened 2015-03-22
Adds configurable network form display

aspic:
I have a use case where the clients don't need an option to set network settings (they are only to connect to the predefined default network), so I omit that from the UI.

Thanks for an awesome IRC web front end! :-)

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -88,41 +88,43 @@
 							<div class="col-sm-12">
 								<h1 class="title">Connect</h1>
 							</div>
-							<div class="col-sm-12">
-								<h2>Network settings</h2>
-							</div>
-							<div class="col-sm-3">
-								<label>Name</label>
-							</div>
-							<div class="col-sm-9">
-								<input class="input" name="name" value="<%= defaults.name %>">
-							</div>
-							<div class="col-sm-3">
-								<label>Server</label>
-							</div>
-							<div class="col-sm-6 col-xs-8">
-								<input class="input" name="host" value="<%= defaults.host %>">
-							</div>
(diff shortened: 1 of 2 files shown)


aspic:
Uh, this didn't work as I intended to, will fix.

erming:
Good idea!

This commit looks pretty straight-forward. Is it working yet? I'll merge it when you respond.

aspic:
Yup, should be working now. :-)

erming:
Thanks!

bnvk:
@aspic I'm curious about your use case as I'm implementing a similar thing for Channels... would it make sense to hide the "X" button in the sidebar & disallow the `/connect` and `/quit` commands as well? Curious @erming thoughts on this as well!

aspic:
@bnvk the use case was to make it easier for users to join one predefined channel for one network (seems similar to yours). :-) Maybe this could be generalized to a "single channel" setting instead. However I don't see the benefit of limiting the user that much, they don't need to join other channels unless they want to ;-)

bnvk:
@aspic cool, glad I'm not the only person seeing this sort of need from users. I'll maybe work on merging our cases in some sensible way once I get a little further along ;)

To elaborate on why I think disabling ability to join / leave channels is better UX. Consider an org that harbors 10 projects / communities (each has its own IRC chan) and the org wants to allow the open pulblic web & regular contributors to participate easily. 

I guess I'm thinking about it, maybe the *more* ideal experience is disallowing "leaving" of set channels but still allowing joining. Main reason being, once you let people delete / hide a channel you want there as org you have to provide easy way (non text commands) to add them back! 

What do you think @jancborchardt ?

Issue #206, opened 2014-10-02
Unread indicator


YuukanOO:
Could not find a valid title for this feature. It would be cool to have an unread indicator when you switch browser tab or channel. Currently when you are on another channel, you could see the number of unread messages but when you go into this channel, you don't know where you left the discussion.

Is it something feasable ?

PR #382, opened 2015-03-24
Add basic unread marker

TobiasBales:
Showing a horizontal line where on left off/started reading.
The last one is kept and no fancy hide after scrolling into and then out of view because I find those highly annoying.
It also works when it is being rendered upon page load

This should address #206


erming:
Thanks!

I will test this after work today. I'll get back to you.

erming:
Hey! I've tested the pull request.

Unfortunately, I ran into some minor errors.

For example, this:
hr

It gives me two unread markers in the same channel after I write something. Also, as you can see in the image, you've accidentally used `<br>` instead of `<hr>` in the template, leaving a small gap on the left side.

Some other minor nitpicks (don't worry, they're not major):
- I usually put Handlebars helpers (except `partial`) in this folder: https://github.com/erming/shout/tree/master/client/js/libs/handlebars
- You've also added an .editorconfig file, but the CSS you've committed still has the wrong indentation. Plus, I think this file should've been a commit on its own.

TobiasBales:
Good pickups, will fix them and report back ;)

astorije:
Hi @TobiasBales, thanks for your PR!

It seems like @erming got too busy after his initial review so I'll take it from here. This is a cool feature and that would be nice to see it in shout.

A few things however:

- Could you remove the `.editorconfig` thingy? There is a pending PR (#504) that takes care of this. I need to get through it but I should be able to spend time on this soon.

- When joining a channel, the line gets stuck to the top, the result is not very pretty: 
double unread line

- Actually, it took me a while to make that feature work, as the line would stay at the top at all time. It seems like it doesn't work in some cases, not sure what these are. Could you test your feature a bit more and see if you can identify flaws? In addition, a new line sometimes gets added when I enter a message (the "shout-user" was me and the bottom line appeared when I entered the message):
screen shot 2015-10-15 at 20 36 13

- It seems to me (I haven't read the code yet) that this line gets added when someone posts a message in a channel which is not active. I would expect a non-focused browser tab to be considered as such: after all, whether I haven't clicked on the channel or switched on the tab, I haven't seen these messages! Could you add/fix this?

astorije:
Also, I would start by rebasing on master, as a lot of things have changed recently.

Xe:
https://github.com/Shuo-IRC/Shuo/commit/5915e960d6062914565b44239406bebc1089647d did the trick on top of this patch for us.

astorije:
Thanks for the useful pointer, @Xe!

Issue #361, opened 2015-02-11
Feature Request: Click on Channel Links


shadowfacts:
Suggestion:
If someone types a channel name in chat, the user should be able to click on it from the web interface and automatically join the channel. e.g. If someone typed #minecraftforge, on the web client #minecraftforge would appear as a link, and when clicked on, shout would join #minecraftforge for them.

erming:
Agree!

This shouldn't be too hard to implement. I'm already scanning every message for links. It could also scan for `#chan` and make them clickable.

shadowfacts:
Sounds Good :smile:

xPaw:
This was fixed.

astorije:
Indeed, fixed in #385, thanks for the triaging help! :-)

PR #385, opened 2015-03-28
made channel names in chat clickable; lets users join channels

AmShaegar13:
Implements #361.

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -608,6 +608,12 @@ button {
 	display: none;
 	font-style: normal;
 }
+#chat .msg .inline-channel {
+	cursor: pointer;
+}
+#chat .msg .inline-channel:hover {
+	opacity: .6;
+}
 #chat .time,
 #chat .from,
 #chat .text {
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -190,12 +190,38 @@ $(function() {
 		var chan = chat.find(target);
 		var from = data.msg.from;
 
+		var msg = $(render("msg", {messages: [data.msg]}));
 		chan.find(".messages")
-			.append(render("msg", {messages: [data.msg]}))
+			.append(msg)
 			.trigger("msg", [
 				target,
 				data.msg
 			]);
+
+		var text = msg.find(".text");
+		if (text.find("i").size() === 1) {
+			text = text.find("i");
+		}
+		// Channels names are strings (beginning with a '&' or '#' character)
+		// of length up to 200 characters.
+		// See https://tools.ietf.org/html/rfc1459#section-1.3
+		text.html(text.html().replace(/(^|\s)([#&][^\x07\x2C\s]{0,199})/ig,
+				'$1<span class="inline-channel" role="button" tabindex="0" data-chan="$2">$2</span>'));
+		text.find("span.inline-channel")
+			.on("click", function() {
+				var chan = $(".network")
+					.find(".chan.active")
+					.parent(".network")
+					.find(".chan[data-title='" + $(this).data("chan") + "']");
+				if (chan.size() === 1) {
+					chan.click();
+				} else {
+					socket.emit("input", {
+						target: chat.data("id"),
+						text: "/join " + $(this).data("chan")
+					});
+				}
+			});
 
 		if (!chan.hasClass("channel")) {
 			return;


client/css/style.css
```diff
@@ -573,6 +573,7 @@ button {
 }
 #chat .wrap,
 #chat .text a {
+	cursor: pointer;
```
astorije: Is it really necessary to add a cursor pointer to `a` links as opposed to default behavior?

astorije:
I'll try to give a review to this PR. Will look for @erming or @JocelynDelalande's eyes as well.

client/css/style.css
AmShaegar13: Well, what means *necessary*. If you want the links to appear and behave the same as normal html links, then yes. It is necessary. If you are satified with the default behavior, which means the cursor changes to `text`, then it is not necessary.

I like the first option.

astorije: Sorry, I didn't see the `#chat .wrap` on the first line. However, I'm confused, all links are wrapped into `a` links (what you called normal HTML links). Do you have an example where we need this?

AmShaegar13: MIght be that one of us is missing something. My point is, when I create a link like so: `<a data-chan="#channel">` it will not have the pointer cursor due to the missing `href`. At least firefox behaves like that. To solve that I added `cursor: pointer;`.

client/js/shout.js
```diff
@@ -197,6 +197,23 @@ $(function() {
 				data.msg
 			]);
 
+		chan.find(".messages .text:not(.processed)")
+			.each(function() {
+				var html = $(this).html();
+				$(this)
+					.addClass("processed")
+					.html(html.replace(/(\W)(#[\w\d-.]{0,200})/ig,
```
Xe: This will match channels longer than the IRC daemon likely supports. Does this need to be 200?

AmShaegar13: I searched for the specification when creating the regular expression and came up with this:

>  Channels names are strings (beginning with a '&' or '#' character) of
>  length up to 200 characters.


https://tools.ietf.org/html/rfc1459#section-1.3

client/css/style.css
astorije: Ah, very good point! I had forgotten about this oddity. Well, the HTML5 spec does say that [the `a` element, absent an `href`, is not an hyperlink but a placeholder](http://www.w3.org/TR/html5/text-level-semantics.html#the-a-element).

[This Stack Overflow](http://stackoverflow.com/a/11145021/1935861) is very well built and even links to WAI-ARIA!

My conclusion on this (also after reading [this](http://css-tricks.com/when-and-when-not-to-use-an-anchor-tag/)) is that we should not go for an `a` link. We should either go for a `button` or a `span` with the right ARIA role and focus behavior. At the moment links on the left menu are using a `span`, links on the right menu are using a `button`. Please go for a `button` and if you realize it's hell to style (not sure how it behaves with long statements that go to a second line) then switch back to a `span`:

```html
<button role="button" tabindex="0">...</button>
```

This is definitely the right way to go for in-app "links" and I'll make sure we go for the right route in subsequent choices. And we keep `a` links for actual hyperlinks.

Does that make sense to you? Let me know if you have any questions, and sorry for being slow at first :-)

AmShaegar13: Ah, sounds good to me. Letzte me change that tomorrow.

AmShaegar13: Let*

client/js/shout.js
astorije: Thanks for linking to an RFC :clap: 
Actually, it's so useful to have it there, could you make it a comment shaped like this please:

```js
					// Channels names are strings (beginning with a '&' or '#' character)
					// of length up to 200 characters.
					// See https://tools.ietf.org/html/rfc1459#section-1.3
					.html(...
```

Three other comments:
- I don't see support for `&channel`, as only `#` is present in your regex. Could you add support for this please?
- Even though the RFC doesn't explicitly say it, I don't think  `#` is a legal channel. So maybe minimal length should be `1` instead of `0`
- The RFC says, however, that channel names are up to 200 characters, `#`/`&` included :-) So it should be `199` and not `200` :-)

client/css/style.css
astorije: Awesome! Thanks for being so responsive, I really appreciate it! Talk to you soon.

client/js/shout.js
AmShaegar13: * I couldn't find a network where &channel is working. Are you sure that this is supported by shout's backend?
* Actually, I do think so. Test it on the shout demo network for example. `/join #` works just fine.
* I agree.

astorije: > - I couldn't find a network where &channel is working. Are you sure that this is supported by shout's backend?

Typing `/join <whatever>` in the input [is simply a forward to Slate with `<whatever>` as an argument](https://github.com/erming/shout/blob/98344b2cac540eb5d862e4bf1c53e73a538c9afd/src/plugins/inputs/join.js), which in turn [merely sends `JOIN <whatever>`](https://github.com/slate/slate-irc/blob/6f2c820ee6afd28b935deafae99c1760b14e6341/index.js#L239) to the network, so I don't think there any massaging on that. Actually, it might be nice to prevent trying to join a network that does not respect the RFC, either on Shout or on Slate, but that's for another PR ;-)

Additionally, I asked @deiu as I know he is using Shout on a `&`-enabled server and he confirmed it works fine.

> - Actually, I do think so. Test it on the shout demo network for example. `/join #` works just fine.

You are correct (and it highlighted #483 ^^), so the right interval is `{0,199}` indeed!

richrd:
Sorry for not contributing anything, but I just want to say that I support this feature. I think it's a good addition.

AmShaegar13:
Thanks for your feedback. Added it to the code. Also I made a litte change to the regex to comply with the rfc completely. Originally, it only allowed a-zA-Z0-9. It's now accepting every character except those mentioned in the RFC: space, comma, \x07

client/js/shout.js
```diff
@@ -197,6 +197,26 @@ $(function() {
...
 
+		chan.find(".messages .text:not(.processed)")
+			.each(function() {
+				var html = $(this).html();
+				$(this)
+					.addClass("processed")
+					// Channels names are strings (beginning with a '&' or '#' character)
+					// of length up to 200 characters.
+					// See https://tools.ietf.org/html/rfc1459#section-1.3
+					.html(html.replace(/(\W)([#&][^\x07\x2C\s]{0,199})/ig,
```
astorije: Why did you replace `[\w\d-.]` with `[^\x07\x2C\s]`? What does the last one mean? Seems more cryptic (if you consider regexes aren't cryptic in the first place...)

astorije: @AmShaegar13 answered [here](https://github.com/erming/shout/pull/385#issuecomment-143590344).

AmShaegar13: RFC states:
> the only restriction on a
> channel name is that it may not contain any spaces (' '), a control G
> (^G or ASCII 7), or a comma (',' which is used as a list item
> separator by the protocol)

astorije:
Nice @AmShaegar13!
I am testing this locally (since the code looks nice :) ), and here are my discoveries:

- Style between normal text and channel names acting as buttons are the same. I won't be blocking this PR for that as I'm OK with it, but an alternative I would display them same as links. Let's see what the second reviewer says.

- It works fine when the channel isn't already in the opened channel from the left list, because it makes you join and open the channel in foreground. If you already have joined this channel, nothing happens. This seems like a bug and what I suggest to solve it is to open the already joined channel. Could you do this? (maybe emit a click?)

- Since this is now a `button`, long lines won't cut the channel name but put it on another line:

  

  Second line shows the exact same text without the `#`. I think this should be consistent, the channel name should not be on its own line. Could you try to see if you can have `button` to do that? If not because of the way `button` is, then feel free to go for a `span` instead ;-)

- And by the way, opening the channel above...

  

  Ouch, can you fix what's going on with that trailing `</i>` please?

AmShaegar13:
I discovered another downside of the button: You cannot select it as text. We should really use span... This should fix long channel names.

`<i class="wrap"></i>` is added around long messages. Didn't see that... The regex matches the closing </i> now... Otherwise it would always be a space or new line.

This took me to another problem: HTML entities are escaped. So instead of joining #channel<s> you would join #channel&lt;s&gt;... Looks like the HTML context is not good for this feature... We may have to rethink where this code belongs :confused:

astorije:
> I discovered another downside of the button: You cannot select it as text. We should really use span... This should fix long channel names.

Yes, go for it :-)

> The regex matches the closing now...

You probably didn't push this, did you?

> Looks like the HTML context is not good for this feature... We may have to rethink where this code belongs

I don't think we should: this definitely belongs to the client, not the server! Plus, in the `msg` incoming from the socket is also fine too.
However, I do realize some problems indeed. That `processed` thing is odd and we shouldn't need it: we are actually processing one message, so let's add the `span` button here only, not trying to loop through all non-processed lines (which should be just this one!).
There are solutions to the HTML tag issue, but we must be careful as it could add security concerns with XSS if not correctly handled. I'll sleep on it and let you fix the rest in the meantime. Thanks!

AmShaegar13:
Here we go. First of all, I replaced button by span. I had to add the rule for the pointer cursor again now. Secondly, I cleaned up the 'processed' thing. I have no ide why I did this in the first place... I made the new message a jQuery object and used this for further processing.

I fixed the </i> issue by working directly on the <i> element, if there is one.

Also, I broke travis-ci, although I'm not sure if it's really my fault. The output says something about a timeout. I can still fet link information on my development instance.

Please let me know what you think!

JocelynDelalande:
> Also, I broke travis-ci, although I'm not sure if it's really my fault. The output says something about a timeout. I can still fetch link information on my development instance.

You didn't :)

Actually, it was broken already, now it's fixed, see #493 please update your PR rebasing on `master`, tests should be passing then.

astorije:
Hi @AmShaegar13, this looks very good, exciting!
As you can see, I left a few inline comments. They are basically only style-related at this stage, meaning we are very close :-)

Trying the changes in a local clone, here is what I notice though:
- There is no style applied when `:hover`ing the channel names. Can you add the opacity thingy that is somewhat consistent over the UI?
- I tried to test the multiline channel names with `#very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-long-channel` but on Freenode it doesn't work. Shorter names that expand on more than one line do work fine, so maybe it's a limitation from the Freenode server?!

As @JocelynDelalande just said, don't worry about the tests, just rebase and you'll be fine :-)

JocelynDelalande:
> I tried to test the multiline channel names with #very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-very-long-channel but on Freenode it doesn't work. Shorter names that expand on more than one line do work fine, so maybe it's a limitation from the Freenode server?!

if more than 50 chars [it's not an irc chan name](https://tools.ietf.org/html/rfc2811#section-2.1).

astorije:
Waouh! @AmShaegar13 and I got tricked by [an older RFC]! Thanks @JocelynDelalande!

@AmShaegar13, could you update your regex to accept the 4 prefixes followed by 49 characters?

Xe:
Test

Xe:
@JocelynDelalande that RFC is near universally rejected by IRC daemon developers.

astorije:
> @JocelynDelalande that RFC is near universally rejected by IRC daemon developers.

This is one of these very strong statements that could really use a reference.

Fact is, it seems to be true on Freenode, at least... (or shout itself refuses to join, which is a different but still relevant problem)

AmShaegar13:
Okay, implemended your feedback but there are still two little issues I just found out.

* The <i> that is wrapped around long lines has no spaces in front of the text. Thus, channels at the beginning of a line do not match anymore
* Links that contain channel-like anchors break (http://example.com/#head)

AmShaegar13:
Okay, fixed that too. The only restriction now is that #channels need to be enclosed by spaces or be at the beginning or end of a line.

JocelynDelalande:
@AmShaegar13 cool :) just tested :

On feature:

- if more than 50 chars, it does nothing if clicked, but still is transformed into link, which is confusing.

On style:

- could you create a constants file that will centralize things like max topic length ? (and their justification, like you did in comment mentioning RFC)
- avoid fixing style on lines you don't touch to change their content, it's confusing for review reading git history (at least put these changes in dedicated commit), no big deal though

> Do you want me to squash all commits and rebase on master after the final review?

Yes please :) it seems to me that they could be one commit.

AmShaegar13:
> if more than 50 chars, it does nothing if clicked, but still is transformed into link, which is confusing.

Well, looks like this depends on the network/irc daemon used. The one I am testing on just drops all characters after the 32nd one. So it will always work. If I set the limit down to 50, it will not work in networks that allow 200 characters. So I'd like to stick with that.

> could you create a constants file that will centralize things like max topic length ? (and their justification, like you did in comment mentioning RFC)

I would rather not insert a constant in a regular expression. It makes it even harder to read and understand.

> avoid fixing style on lines you don't touch to change their content, it's confusing for review reading git history (at least put these changes in dedicated commit), no big deal though

Sorry, just used search an replace ^^

JocelynDelalande:
> Well, looks like this depends on the network/irc daemon used. The one I am testing on just drops all characters after the 32nd one. So it will always work. If I set the limit down to 50, it will not work in networks that allow 200 characters. So I'd like to stick with that.

Do you know any network allowing 50+ chars in chan-name ? (actual question, no idea of the answer)

AmShaegar13:
https://www.quakenet.org/ does.

JocelynDelalande:
> https://www.quakenet.org/ does.

Ok, so why not… :+1:  :)

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+		// See https://tools.ietf.org/html/rfc1459#section-1.3
+		text.html(text.html().replace(/(^|\s)([#&][^\x07\x2C\s]{0,199})/ig,
+				'$1<span class="inline-channel" role="button" tabindex="0" data-chan="$2">$2</span>'));
+		text.find("span.inline-channel")
+			.on("click", function() {
+				var chan = $('.network')
+					.find('.chan.active')
+					.parent('.network')
+					.find(".chan[data-title='" + $(this).data("chan") + "']");
+				if (chan.size() == 1) {
```
astorije: Can you replace with `===` here please?

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
 		chan.find(".messages")
-			.append(render("msg", {messages: [data.msg]}))
+			.append(msg)
 			.trigger("msg", [
 				target,
 				data.msg
 			]);
 
+		var text = msg.find('.text');
+		if (text.find('i').size() == 1) {
```
astorije: Can you replace with `===` please?

astorije:
> > could you create a constants file that will centralize things like max topic length ? (and their justification, like you did in comment mentioning RFC)

> I would rather not insert a constant in a regular expression. It makes it even harder to read and understand.

Agreed, and I'm also against centrally-managed constant files here. Many reasons to that but I have seen that in practice and it didn't work well. Simple is enough, and in the future, I'd like to see this very feature being a plugin (but valued as core feature and enabled by default) so having self-sufficient code is ideal.

I'll give a full final review whenever I can find time to do so, sorry for the delay.

client/css/style.css
```diff
@@ -608,6 +608,12 @@ button {
 	display: none;
 	font-style: normal;
 }
+#chat .msg .inline-channel {
+	cursor: pointer;
+}
+#chat .msg .inline-channel:hover {
+	opacity: .8;
```
astorije: I noticed existing opacities are floating between `.6` and `.8` but here I'd prefer `.6` to be consistent with existing links on the chat window.

astorije:
Interesting, when you click on the channel name (in Chrome), you get a link border for a second, but not on the nick names. I guess that's one difference between buttons and spans here...

astorije:
Apart from the minor comments I left inline, this looks good to me. Will wait for final amend before :+1:.

Getting there, getting there! :-)

AmShaegar13:
I am sorry. I was very busy. Finally included you remaining points.

astorije:
@AmShaegar13 thanks! To be honest, I had forgotten about this, but hopefully this is almooost ready to go. If you take a look at the Travis report, you'll see that your PR breaks a styling rule. Could you use double quotes instead of single quotes? I'll be happy to merge after that!

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+		var msg = $(render("msg", {messages: [data.msg]}));
 		chan.find(".messages")
-			.append(render("msg", {messages: [data.msg]}))
+			.append(msg)
 			.trigger("msg", [
 				target,
 				data.msg
 			]);
 
+		var text = msg.find('.text');
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+		if (text.find('i').size() === 1) {
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+			text = text.find('i');
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+		}
+		// Channels names are strings (beginning with a '&' or '#' character)
+		// of length up to 200 characters.
+		// See https://tools.ietf.org/html/rfc1459#section-1.3
+		text.html(text.html().replace(/(^|\s)([#&][^\x07\x2C\s]{0,199})/ig,
+				'$1<span class="inline-channel" role="button" tabindex="0" data-chan="$2">$2</span>'));
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+		text.find("span.inline-channel")
+			.on("click", function() {
+				var chan = $('.network')
+					.find('.chan.active')
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+					.parent('.network')
```
astorije: Use double quotes here.

client/js/shout.js
```diff
@@ -190,13 +190,39 @@ $(function() {
...
+			text = text.find('i');
+		}
+		// Channels names are strings (beginning with a '&' or '#' character)
+		// of length up to 200 characters.
+		// See https://tools.ietf.org/html/rfc1459#section-1.3
+		text.html(text.html().replace(/(^|\s)([#&][^\x07\x2C\s]{0,199})/ig,
+				'$1<span class="inline-channel" role="button" tabindex="0" data-chan="$2">$2</span>'));
+		text.find("span.inline-channel")
+			.on("click", function() {
+				var chan = $('.network')
```
astorije: Use double quotes here.

client/js/shout.js
AmShaegar13: Really? That would require either escaping all the double quotes in the string or replacing them by single quotes.

JocelynDelalande: @AmShaegar13 good point. #547 got merged , you don't have to fix this one, rebasing will drop the warning for this one.

astorije:
GitHub doesn't notify commits, so we were not notified you rebased :-)

That's a :+1: for me.
@JocelynDelalande?

JocelynDelalande:
:+1: merging, thanks @AmShaegar13 !

astorije:
@JocelynDelalande, you closed instead of merging :-)
Merging now.

Issue #295, opened 2014-11-20
Provide the log feature to public (anonymous) users


philippejadin:
We'd like to install shout for apertus.org. (see irc.apertus.org and https://apertus.org/irc-chat) 

Our use would be to allow new users to be up to speed with the project faster. 

So if any visitor could visit the irc client and connect to our channel and have the client load previous logs (let's say latests 500 lines for example), it would bring a lot of interaction and participation, even for new comers, without the need to create an account for everyone.

This is linked to #251

PR #391, opened 2015-04-04
Loads backlog from filesystem if logging enabled.

aynik:
This patch aims to reduce the memory footprint of the backlog. Before it, messages were just accumulated on memory. This adds now the possibility to load messages from the filesystem by seeking and parsing log files. This functionality is only triggered when logging is activated, keeping around 200 messages in memory and loading more if necessary. This modification came out of necessity as I run shout in a very limited VPS with just 128M of memory.

PD. Please consider auto-replacing tabs for spaces in the original code.

PR #392, opened 2015-04-05
Enable logging in public mode and load messages on connect

jdve:
Always log messages, even in public mode, as in #295.  Messages will only be logged as long as at least one public client is connected to a channel.  When all clients disconnect, messages aren't logged anymore.

Also, load messages from the log file when joining a channel.

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -179,6 +179,15 @@ $(function() {
 		}
 		whois = false;
 		chan.click();
+
+		if (data.messages) {
+			var messages = chan.find(".messages");
+			messages.scrollTop(messages.scrollHeight());
+
+			if (data.chan.messages.length == 100) {
+				chan.find(".show-more").addClass("show");
+			}
+		}
 	});
 
 	socket.on("msg", function(data) {
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "read": "^1.0.5",
     "request": "^2.51.0",
     "slate-irc": "~0.7.3",
-    "socket.io": "~1.0.6"
+    "socket.io": "~1.0.6",
+    "ttl": "~1.3.0"
   },
   "devDependencies": {
     "grunt": "~0.4.5",
--- a/src/client.js
+++ b/src/client.js
@@ -79,21 +79,19 @@ Client.prototype.emit = function(event, data) {
 		this.sockets.in(this.id).emit(event, data);
 	}
 	var config = this.config || {};
-	if (config.log === true) {
-		if (event == "msg") {
-			var target = this.find(data.chan);
-			if (target) {
-				var chan = target.chan.name;
-				if (target.chan.type == Chan.Type.LOBBY) {
-					chan = target.network.host;
-				}
-				log.write(
-					this.name,
-					target.network.host,
-					chan,
-					data.msg
-				);
+	if (event == "msg") {
+		var target = this.find(data.chan);
+		if (target) {
+			var chan = target.chan.name;
+			if (target.chan.type == Chan.Type.LOBBY) {
+				chan = target.network.host;
 			}
+			log.write(
+				this.name,
+				target.network.host,
+				chan,
+				data.msg
+			);
 		}
 	}
 };
--- a/src/log.js
+++ b/src/log.js
@@ -1,11 +1,72 @@
+var crypto = require("crypto");
 var fs = require("fs");
 var mkdirp = require("mkdirp");
 var moment = require("moment");
+var Cache = require("ttl");
 var Helper = require("./helper");
+var Msg = require("./models/msg");
+
+var recent = new Cache({ ttl: 10 * 1000 });
+
+function getLogPath(user, network, chan) {
+	return Helper.HOME + "/logs/" + user + "/" + network;
+}
+
+function getLogFilename(chan) {
+	return chan + ".log";
+}
+
+module.exports.load = function(nick, user, network, chan, callback) {
+	var filename = getLogPath(user, network, chan) + "/" + getLogFilename(chan);
+
+	fs.readFile(filename, function(err, data) {
+		if (err) {
+			callback();
+			return;
+		}
+
+		var config = Helper.getConfig();
+		var format = (config.logs || {}).format || "YYYY-MM-DD HH:mm:ss";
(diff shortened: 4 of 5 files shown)


furier:
:+1:

aynik:
I think we may be interested in joining efforts here, I've submitted a pull request (#391) which adds log file parsing too. Are you parsing events too or just messages?

jdve:
Sure!  I'm only parsing messages right now, but I'm sure my changes could fit on top of yours.

astorije:
Hi there,

Quick question: what are the benefits of logging public instances?
In particular, assuming we display scrollback based on logs (/!\ not yet implemented), if someone logs with username `foobar`, logs out, and someone else logs with username `foobar`, does that mean the second user will be able to look at the first user's logs?!

furier:
@astorije good question, I would guess this would only be for public channels "you" enter. Tho logging in public mode without any login makes no sence to me...

astorije:
I'm really confused by this PR. Why would you log public instances?
The only reason I can find is if Shout crashes, and I don't think this PR addresses such use case.
Also, it seems like this PR is adding their own favored in-memory cache while there is already such mechanism for private instances. Why using a new dependency?

I don't think I'm going to give a thorough review to this until @jdve comes back here (or on the IRC channel) because these are important questions to answer before considering a merge. I also invite @JocelynDelalande to come give his input here if he finds the time.

JocelynDelalande:
@astorije benefit is : when you join a channel, you get the recent history of the channel (so it gives you context), I experimented it with [yunohost jabber chan](http://yunohost.org), it's quite handy.

On the other hand, I think enabling this by default for all chans is quite dangerous from a privacy standpoint.

In fact, with your patch, stop me if I'm wrong @jdve (haven't tested, just read), but if I log to the same shout server as you, I will even recover the history from your private chats (queries).

The privacy issue is huge with queries, but also present with public chans where a user may not thinks he is logging publicly the chan to the world.

IMHO such a feature is acceptable only if : 

- it's disabled for queries
- it's enabled by shout admin on a per-chan basis

astorije:
> it's enabled by shout admin on a per-chan basis

Even that is a big problem privacy-wise IMO. When I was at W3C for example (on W3C IRC server), we were often "gathering" in small rooms and expecting not to be logged.

Usually, logging public channel is taken care of by owners of the channel, through a bot, and they often advertise in the topic that the channel is being logged.

I still think that logging on the shout server for public instances could be very harmful. Essentially, I could imagine myself asking people if they are using Shout operated by someone else than themselves before speaking freely.

JocelynDelalande:
> it's enabled by shout admin on a per-chan basis
> 
> Even that is a big problem privacy-wise IMO. When I was at W3C for example (on W3C IRC server), we were often "gathering" in small rooms and expecting not to be logged.

Yep, but using something like shout, you are already in a situation where you trust shout admin server not to deliberately spy on you. The threat model don't really change.

> Usually, logging public channel is taken care of by owners of the channel, through a bot, and they often advertise in the topic that the channel is being logged.

Well… I know several channels which are being logged by individuals buddies, without specific rights, and sometime published , because they consider it a public place.

Maybe display some icon when it's a logged channel ?

astorije:
> Yep, but using something like shout, you are already in a situation where you trust shout admin server not to deliberately spy on you. The threat model don't really change.

Yet again I am all opposed to designing a system that expects everyone to be super nice. We have seen too much in the past that some people are actually using these assumptions to do mass surveillance. There shouldn't be a reason to avoid designing secure systems, even in a situation where you trust your sysadmin.

> Well… I know several channels which are being logged by individuals buddies, without specific rights, and sometime published , because they consider it a public place.

Correct, but again that's very different. If you discuss on a public channel with dozens of users you don't know, you pay attention to what you say. If you are creating a "private" (as in not sharing the channel name with others) channel between you and your parents (or a couple of colleagues or whatever) about sensitive details and you know these ones are trustworthy enough when you wonder if they are logging this channel, do you really expect the admin to log what you say? I clearly wouldn't.

> Maybe display some icon when it's a logged channel ?

That might be a decent minimal requirement for this, but I am still not entirely sold on the matter...

JocelynDelalande:
> it's enabled by shout admin on a per-chan basis

@astorije I'm 
not sure you replied about the per-chan thing, if the adminsys deliberately logs a channel, that's a volunteer and targeted "spying", it's easy to do it by other means (eg HTTP logging), and preventing that is out of our scope IMHO. It's not "mass" and by default logging, which is, I agree with you, to avoid.

astorije:
> @astorije I'm not sure you replied about the per-chan thing, if the adminsys deliberately logs a channel, that's a volunteer and targeted "spying", it's easy to do it by other means (eg HTTP logging), and preventing that is out of our scope IMHO. It's not "mass" and by default logging, which is, I agree with you, to avoid.

I [did answer](https://github.com/erming/shout/pull/392#issuecomment-143890993).
Anyway, I disagree with that comment, but I am done explaining why. Feel free to self-assign that PR if you want to follow up with it.

Xe:
Who's to say the irc server administrator is not also spying on you?

You gotta trust someone somewhere eventually. That's just how stuff works.

floogulinc:
I honestly don't see much of a point in this feature and I think it has too many possible problems. Most people would expect the current behavior when in public mode, to basically act as a normal web IRC client.

JocelynDelalande:
Ok, trying to sum up things before it turns into troll (feel free to amend if it seems not to reflect the debate).

## Proposed features

(as submitted by @jdve)

- log all messages in public modes (query & public chans)
- load the logs on start

## Pros

- gives context when you join a chan
- simple and natular access to chan backlog

## Cons (without modifications)

- I can pretend to be any user and gets its query (or small "private" chans) log
- Can be used as per-default "spying" tool by adminsys and anyway

## Proposed modifications

- do not log anything by default
- adminsys can enable per-chan logging
- display an icon to warn user on a logged chan (it only warns local shout users, not other users)

## Cons (with proposed modifications)

- could still be used by shout admins to target some chan logging
- may be an unexpected behaviour

JocelynDelalande:
## Proposal
As :
- there is a disensus on that PR even with modifications
- @jdve seems no longer active on this

Let's reject and close this PR for now.

If one day we want change our mind, we will have the debate summary I did just before as a starting point.

What do you think ?

PR #397, opened 2015-04-14
Crypto theme

aynik:
Added new theme 'crypto'. Attached fonts (Leto and Droid Sans Mono) to use them without hitting google.

login
config
chat


tracker1:
+1, though I'm more fond of Inconsolata-g as a fixed-width font over Droid Sans Mono.

FYI: If "Leto Black" is intended to be used as a bold "Leto", don't separate "Leto" and "Leto Black" just assign the appropriate `weight` to your `font-face` declarations.  This way you can use combinations of the `Leto` family along with `bold` ...

aynik:
@tracker1 I've done what you suggested but I find Inconsolata-g a little too big, I've reduced it on chat messages to a "non-standard" size of 11px to look reasonably good.

erming:
This theme looks sweet! Reminds me of http://hackertyper.com/, haha!

I just tested it out and I really like the looks. I think I'll use this as my default theme. However, there are a couple of tiny fixes that I will fix in another commit in a few minutes:
- There seems to be a gap above the messages and sidebar
- You changed the default font in style.css. This breaks the default theme a bit, so I'm making this font specific to your theme instead and then roll back the default to Consolas.

I also appreciate that you added the fonts locally instead of Shout relying on Google Fonts. I should've done it like this from the start.

Thanks man!

aynik:
cool! I'll pull when you're ready with it.

erming:
@aynik, just a quick question:
Why Lato-Black instead of Lato Bold? See http://fonts.googleapis.com/css?family=Lato:400,700

Also, the Lato you provided seems to render a bit differently:

lato_aynik

Compare the first input and you can see the different line heights. There's also a slightly difference how they both render, but that's just a minor thing. Where did you get your `lato.woff` from?

aynik:
@erming It is indeed a newer version, I got it from [here](http://www.latofonts.com/lato-free-fonts/). The site says:

> The older ver­sion (1.0) of the Lato font fam­ily is avail­able on Google Fonts. Using this ser­vice, you can eas­ily incor­po­rate the Lato fonts into your web­site. (In fact, this very web­site uses the Lato fonts served by Google Fonts). Soon, the new 2.0 ver­sion will be avail­able on Google Fonts as well.

erming:
Ok, I did a rollback on the font (now uses Google Fonts again). I will eventually add them locally again.

Anyhow, new version is out now:
https://github.com/erming/shout/commit/7138ca6bda5430d1fdbc9f6e530d9cfd11247bd1

Thanks again!

PR #402, opened 2015-04-24
User-specific bind address setting

kalaksi:
Because many IRC servers allow only a few connections from the same IP address, it becomes necessary to use multiple IP addresses for IRC connections. So I added a setting for binding users' connections to different addresses.

Btw, without setting server.family I couldn't get binding to IPv6 addresses to work.
I haven't tested if setting server.family in nodejs 0.10 causes problems (or if anyone cares about 0.10?) but I think it shouldn't. I can test it if no one confirms.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -67,6 +67,8 @@ function Client(sockets, name, config) {
 		var delay = 0;
 		(config.networks || []).forEach(function(n) {
 			setTimeout(function() {
+				// User-specific bind (localAddress) for networks
+				n.bind = config.bind;
 				client.connect(n);
 			}, delay);
 			delay += 1000;
@@ -126,15 +128,16 @@ Client.prototype.connect = function(args) {
 		name: args.name || "",
 		host: args.host || "irc.freenode.org",
 		port: args.port || (args.tls ? 6697 : 6667),
+		localAddress: args.bind || config.bind || undefined,
 		rejectUnauthorized: false
 	};
+	
+	// XXX: "family" requires nodejs 0.12, but setting it won't hurt
+	server.family = net.isIP(server.localAddress) || undefined;
 
-	if (config.bind) {
-		server.localAddress = config.bind;
-		if(args.tls) {
-			var socket = net.connect(server);
-			server.socket = socket;
-		}
+	if (server.localAddress && args.tls) {
+		var socket = net.connect(server);
+		server.socket = socket;
 	}
 
 	var stream = args.tls ? tls.connect(server) : net.connect(server);
--- a/src/clientManager.js
+++ b/src/clientManager.js
@@ -80,6 +80,7 @@ ClientManager.prototype.addUser = function(name, password) {
 			user: name,
 			password: password || "",
 			log: false,
+			bind: "",
 			networks: []
 		};
 		mkdirp.sync(path);


JocelynDelalande:
Hi @kalaksi, happy to speak with the you from by 6 months ago ;)

> Btw, without setting server.family I couldn't get binding to IPv6 addresses to work.
> I haven't tested if setting server.family in nodejs 0.10 causes problems (or if anyone cares about 0.10?) but I think it shouldn't. I can test it if no one confirms.

Please do, ubuntu (14.04) and Debian (jessie) do come with 0.10 and 0.10 it's in the versions list we list in the [.travis.yml](https://github.com/erming/shout/blob/master/.travis.yml).

Added to the fact it seems a bit hackish to me, I have some concerns about that feature though : does it means a shout user can bind to the address he wants ? (crafting the request to modify the settings).

kalaksi:
I didn't think about that, it is possible if shout allows user to arbitrarily define the variables in user's config file. IMO, there should be a list of options that the user is allowed to change, since this one probably won't be the only one that should be restricted (e.g. ident-names, too). 
This feature doesn't make the bind-option mandatory, though.

Having a dedicated addresses for users is what irccloud.com seems to do: they have both identd-name and dedicated address such as uidNNNN@id-NNNN.AAAA.irccloud.com. This is also useful for some basic identification.

JocelynDelalande:
@kalaksi so merging as-is isn't safe

Isn't webirc (see #379 ) the propper way to handle the issue you are trying to circumvent here ? (just asking, really, to be sure that we make the right choices).

kalaksi:
I think it's just a different solution with a different setup. AFAIK WebIRC-client is usually a web client that doesn't have a persistent backend keeping the connections open and the protocol's intention is to forward the user's actual IP address (or at least generate an ident/username from it) and therefore act like the user would be running it locally.

However, user using shout will have a different IP address (dynamic, different place, mobile etc.) and therefore would require a new connection to the IRC servers to have the ident/username and/or hostname to be updated to match the users' current address.

Hostname spoofing also requires changes on IRC server configuration and setting up passwords.

I think shout is more comparable to having a shell provider with a text-based client such as irssi / weechat: the actual IRC client is on the provider's server and differentiating between users happens usually with ident (the shell provider should have an ident server set up). Having dedicated addresses for users would be an alternative and as mentioned, irccloud.com uses both.

JocelynDelalande:
@kalaksi Ok, to sum up things :

- do you know connections IP limitations that are actually blocking you or is it pure speculation ? (I mean limitations in the scope of shout, which is not designed to scale specially well...)

With your suggestion :

- for IPv4, it's not likely you will get a lot of addresses, so mostly useless (excepted very big shout installs, which is not likely to happen, see scalability issue)
- for IPv6, I suppose the connections limitats are at least per /64 or /48, (but that's speculation, I don't know any running IRC network actual rules on that), any data on that

@kalaksi still trying to figure out if it's an interesting feature or a strange hack ;-)

kalaksi:
(I was just about to add that in theory, the webirc-protocol could help with the original issue but would create other problems, mainly the webirc-configuration for different IRC servers (if even available) and privacy issues.)

About the limitations, it's common to see limits such as 3-5 per IPv4 IP. Acquiring multiple IPv4 addresses is probably mostly limited to enterprises and such. I can get 3 addresses at home, though, which in theory allows me to triple the amount of allowed connections.
For me this feature isn't actually useful for the initial purpose anymore, since I want to use IPv6 and IPv6 servers limit the connection count based on a wider prefix (one has limit of 5-10 per /48), so this isn't that important to me anymore. I used to have servers that didn't. It's probably better to request higher limits from the irc admins, anyway :blush:. 
 
I still prefer this kind of setup, though. It allows better separation between users on the network level too and I could think of some uses (other than identification) such as easier debugging (for battling abuse, too, e.g. immediately find out which users have connection to a certain server or how much they consume traffic) or limiting connections.

I don't consider this feature itself a hack, but I know it isn't the most common or the most useful of features. Irccloud uses it :wink:. (And consider this: users on different VMs with screen + irssi allow this kind of setup with dedicated addresses. Or actual computers of different family members at home.)

By the way, the same security issue will rise with #379 IF webirc is implemented for registered users too and the user's IP address is stored in the user config for persistent connections.

astorije:
I am also quite baffled by the problem and its solution here.

Being blocked after 3-5 connections seems hardly believable to me. A lot of small companies who are using IRC as an organization are, I'm sure, behind a single IPv4 address and you don't hear often they get kicked out.

Also, you are saying that `screen` + `irssi` (so I'm guessing in a configuration where N users are SSHing to a single server with a single IP address) does the same, I'm curious, what does it do exactly?
And you are saying that irccloud is doing the same as well, could you take a look at their code doing so?

I think the solution itself is indeed a hack, and relying on people having multiple IPv4 addresses is an assumption I wouldn't make, considering a good bunch of Shout users are self-hosting.

kalaksi:
> Being blocked after 3-5 connections seems hardly believable to me.
For example, QuakeNet: https://www.quakenet.org/news/224-updated-connection-restrictions-per-ip-address

For QuakeNet, users can request "trust" for higher limits, which I have done. 
Similarly, 2 connections per IP for an IRCnet server and no possibility for raising the limit.

> Also, you are saying that...

I'm just saying it allows network level separation. I have previously provided shell access and VMs and had separate IPv6 addresses for users as they are on separate VMs. With IPv6, you will get a prefix with a ton of addresses for use. (And even if on a same server with multiple addresses, you can define the local IP address for the IRC connections in irssi. As with WeeChat, also. It's not that rare.)

> could you take a look at their code doing so

Nope, just /who \*irccloud\*, look at the results and do a lookup to find out that users have separate IPv6 addresses. Well, I looked it up: https://www.irccloud.com/networks 
"If your network supports IPv6 connections, we'll detect any AAAA records exposed in DNS, and favour those by default. This will allow us to assign each user a unique IPv6 address, mapped from their user ID, e.g. 2604:8300:100:200b:6667:1::[UID_IN_HEX]. This is the preferred way to allow IRCCloud support, since each IRCCloud user connects from their own unique address within our /64 range."

The *solution* for the initial problem might be hackish, but not the feature itself. Sometimes there's no alternative and irssi and WeeChat have the feature to let me use separate addresses (or using VMs / containers).

JocelynDelalande:
@kalaksi ok, we're moving forward :) Thanks for bringing more input, actual links and figures you provide do help. I think a part of the difficulty of conversation is that it requires thinking IPv6 actually, which is quite new logic for a lot of people.

Your PR would require IMHO those modifications to become legitimate:

- be sure not to entangle the code too much in core, this will probably move to a plugin if we manage to write a plugin system [one day](https://github.com/erming/shout-roadmap).
- Manual ipv6 binding per-user is not an option: manually specifying arbitrary IP in user config is weird, error-prone and has no value IMHO, excepted to complicate the setup of the feature.
- I'd suggest setting a list/subnet of IPv6 in *config.js* (or even the oportunity to get "all" available and publicly routable addresses), and first time a user connects, we "bind" him/her to that address (register it in user conf, this setting should not be user-editable).
- you have to think about "public" mode for your feature (I don't know, may just be disabled for public mode, but let's think this)
- If I'm not wrong, your PR will define that a user either connects with IPv4 or IPv6, which is not what you want (it should depend on IRC server connectivity options, not on user choice). So address selection should be smarter
- You have to check on shout startup and at IRC connection that the addresses in address list are actually bound to the machine and fail or fallback nicely if not.
- target specifically this PR at IPv6, it will simplify some things (address selection/fallback for example), and I think it's more valuable here (getting several IPv4 is not something everyone can afford, and getting one IPv4 per-user on a shared shout is quite irresponsible regarding [IPv4 exhaustion](https://www.arin.net/announcements/2015/20150924.html) + IPv6 is widely deployed in IRC world, let's take advantage of it.
- Ensure that this user setting is not editable by user

That would turn a personal hack into a feature IMHO. So two questions:

1. do other maintainers (ping @astorije @erming ) think it's worth that amount of code¹
2. @kalaksi are you ready to spend that amount of time writing the code ?

For myself, I'm ok with the feature if implemented correctly.

kalaksi:
@JocelynDelalande good and fair points. At it's current condition this doesn't manage any IP pools automatically. It shouldn't take too much to implement, though. For my uses the intention was to assign the IPs manually. If no assignment isn't done, it uses the globally specified bind address (which, if defined, probably doesn't allow dual-stack either, btw?). This was done to replace screen + irssi -access where registration isn't automatic anyway. 
I had no intention to use dual-stack addresses simultaneously, only IPv6 or IPv4 at a time. As you said the protocol family selection could be made smarter to be more usable generally (same goes for the global IP, too?).

I would have been interested in improving this PR according to your good suggestions, but sadly, I gave up on shout so I have no intention on spending too much time on these PRs anymore. That said, I completely understand if you want to reject this as there probably isn't too much interest for this kind of feature, at this point.

JocelynDelalande:
@kalaksi ok, so it's a :-1: from me to integrate it as-is, closing the PR (could be re-opened if someone wants to take the time to build the feature)

@kalaksi @astorije interesting discussion anyway, thanks for your inputs and for being able to have a long but troll-free constructive discussion; it's pleasant :)

JocelynDelalande:
@kalaksi oh and other question if no impolite why did you gave up with shout ?

kalaksi:
@JocelynDelalande not at all impolite! I'd do the same and ask just for the feedback.
I have long been looking for a replacement for that "screen + irssi"-like combination and I tried shout for some time. These PRs I made reflect the rather small initial fixes that I wanted to implement to make it suitable for my use case. I may have made the PRs too hastily. I didn't use / test shout for that long (had 1 friend use it for some time, too), but at the time shout seemed like the best alternative and I was ready to continue developing these features.

Shout is promising and there sure are many worse alternatives, but I quickly found that shout just isn't that mature yet. I was ready to contribute (for the IRC!) but different factors (some things are just very messy) and inactive development made it look like any changes aren't going to happen anytime soon. (I'm also a noob with github :blush:), but most importantly: I found Quassel. It has clients for different platforms, so I started using it. It has it's own problems, though. 

I took a look at the commit log just now and I'm happy to see that the development has definitely speeded up again! I hope this project makes IRC access easier for many people.

astorije:
Hi @kalaksi,

Thanks for your detailed answer, it does give much more context and helps understanding why this is needed.

As @JocelynDelalande said, this PR is unfortunately not in a good shape to be merged as-is. However, when we manage to have plugin mechanisms, I'd be very happy to see that come back. After reading and re-reading comments from both of you, I think it would be neat to have such a feature on IPv6 addresses.

That being said, I'm very sad to see we cannot count you as a contributor anymore :(
Hopefully the work we are currently doing will help you change your mind someday. As you said, the current codebase makes it really hard to add code that would not be convoluted into other code, as well as not really testable. The plan is to get rid of these limitations, so that coding on Shout is less of a fight!
In the meantime, that would be helpful if you open a ticket on the roadmap (https://github.com/erming/shout-roadmap) to share your past feelings on the project.

PR #409, opened 2015-04-26
Fix topic on UI

kalaksi:
Enables topic again and moves escaping to the client side.

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -314,6 +314,10 @@ $(function() {
 		}
 	});
 
+	socket.on("topic", function(data) {
+		$("#chan-" + data.chan).find(".header .topic").html(data.topic);
+	});
+
 	socket.on("users", function(data) {
 		var users = chat.find("#chan-" + data.chan).find(".users").html(render("user", data));
 		var nicks = [];
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -315,7 +315,11 @@ $(function() {
 	});
 
 	socket.on("topic", function(data) {
-		$("#chan-" + data.chan).find(".header .topic").html(data.topic);
+		// .text() escapes HTML but not quotes. That only matters with text inside attributes.
+		var topic = $("#chan-" + data.chan).find(".header .topic");
+		topic.text(data.topic);
+		// .attr() is safe escape-wise but consider the capabilities of the attribute
+		topic.attr("title", data.topic);
 	});
 
 	socket.on("users", function(data) {
--- a/src/plugins/irc-events/topic.js
+++ b/src/plugins/irc-events/topic.js
@@ -9,27 +9,24 @@ module.exports = function(irc, network) {
 			return;
 		}
 		var from = data.nick || chan.name;
-		var self = false;
-		if (from.toLowerCase() == irc.me.toLowerCase()) {
-			self = true;
-		}
 		var topic = data.topic;
+
 		var msg = new Msg({
 			type: Msg.Type.TOPIC,
 			mode: chan.getMode(from),
 			from: from,
 			text: topic,
-			self: self
+			self: (from.toLowerCase() === irc.me.toLowerCase())
 		});
 		chan.messages.push(msg);
 		client.emit("msg", {
 			chan: chan.id,
 			msg: msg
 		});
-		chan.topic = topic
+		chan.topic = topic;
 		client.emit("topic", {
 			chan: chan.id,
-			topic: _.escape(topic)
+			topic: chan.topic
 		});
 	});
 };
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -403,16 +403,14 @@ button {
 	line-height: 50px !important;
 	height: 48px;
 	padding: 0 20px;
+	overflow: hidden;
 }
 #windows .header .title {
 	font: 14px Lato;
 }
 #windows .header .topic {
-	/* Hidden for now */
-	display: none;
 	color: #777;
 	margin-left: 8px;
-	text-transform: capitalize;
 }
 #windows .header .right {
 	float: right;


astorije:
This seems both like a bug and a feature :-)
I will give it a review soonish, thinking it should not too difficult to get that back (prioritizing easy fixes/features as a first step).

@erming or @JocelynDelalande, although I do want to give a review to this, I'll want a second pair of eyes, so feel free to give your comments as well.

astorije:
@kalaksi, this seems to handle topic change when the current shout user runs a `/topic` command.

- What happens if another user changes the topic (aka the event does not come from the `socket` object but from the `irc` object?
- Is the existing topic correctly handled? That is, if no `/topic` from any side happens, is it still properly rendered?

kalaksi:
I couldn't remember for sure what cases I tested so I just tested this again and it works in all of those cases. I haven't really looked into shout's code for a while but "node_modules/slate-irc/lib/plugins/topic.js" seems to take care of handling the incoming topic-replies.

kalaksi:
Although, I'm not sure where it is escaped on those cases...

kalaksi:
OK, as I understand it now (again), the socket in client/js/shout.js is the socket between shout-client shout-server and as such receives events from the shout-server, not from user input (user-input is handled on line 404). The irc-object you're probably talking about is located on shout-server and not directly related to escaping the HTML client-side.

So, to me it looks like the proper place for escaping, because it is the only place where topic is set in client UI.

astorije:
HTML-specific escaping should always happen at display time anyway, so I'm fine with that. Let me take a closer look to your PR and I'll come back with questions.

In the meantime, before re-enabling any support for this, I am very curious to know why @erming explicitly disabled topic display in the first place. @erming, care to jump in?

client/js/shout.js
```diff
@@ -314,6 +314,11 @@ $(function() {
 		}
 	});
 
+	socket.on("topic", function(data) {
+		// .text() escapes HTML (but not quotes)
```
astorije: > `(but not quotes)`

How is that relevant to us here? We need to escape HTML to avoid XSS, but it seems to me that escaping quote would be only relevant to avoid SQL injection, which is not an issue here.

src/plugins/irc-events/topic.js
```diff
@@ -9,11 +9,9 @@ module.exports = function(irc, network) {
 			return;
 		}
 		var from = data.nick || chan.name;
-		var self = false;
-		if (from.toLowerCase() == irc.me.toLowerCase()) {
-			self = true;
-		}
+		var self = (from.toLowerCase() == irc.me.toLowerCase());
```
astorije: Looks like someone has a style preference here :P

So, since you want to play that game, let me be picky ^^

- **Always** use `===` over `==`. There are effectively no good reasons to use the latter over the former (apart from saving one character or using the side-effects of `==` on purpose, none of which are good reasons) but there are plenty good reasons to use the former over the latter.
- Because of the way JavaScript treats `this`, it is extremely common to see functions starting with `var self = this;`. Therefore, `self` is generally reserved for this purpose (along with `that`, but it is frankly less common). See http://jscs.info/rule/safeContextKeyword for example. Let's keep our codebase future-proof when world-class JS developers jump in with their habits :-)

In conclusion, add to this that `self` is just used once, please move this block to line 22 such that:

```js
  self: from.toLowerCase() === irc.me.toLowerCase()
```

(NB: @JocelynDelalande [DRYed this up](https://github.com/astorije/shout/commit/6165795d7c90d80f9be24ce2f7406234879be398#diff-4b7d8172f3d87ca98fe952995318443aR61), but it will be for later when it makes it to master)

src/plugins/irc-events/topic.js
```diff
@@ -26,10 +24,10 @@ module.exports = function(irc, network) {
 			chan: chan.id,
 			msg: msg
 		});
-		chan.topic = topic
+		chan.topic = topic;
```
astorije: Can you or @erming remind me why we need to update `chan`'s `topic`?

client/js/shout.js
kalaksi: At first I wondered too why I had included that comment, but it's probably a sidenote for a case when applied text wouldn't be inside tags but inside an attribute where quotes would matter.

src/plugins/irc-events/topic.js
kalaksi: About the style: I wasn't trying to specifically change the style (or play any games...), but I think that was more about simplifying the code. I know I should have used '===', but either didn't notice it or think about it at the time and I was trying to keep the changes to a minimum. Even though this isn't strictly speaking on the scope of fixing the topic, I thought a small simplification wouldn't hurt anyone.

astorije:
@kalaksi I tested locally and it's pretty cool! I have a few comments though:

- Words get uppercased, I must admit it's the first time I see that. Example, when submitting `/topic Test <b>topic</b>`, I get:

  topic

  Urgh. I know this is not your addition that did that, but since we are on the topic topic :-)
Could you make sure that whatever one types after `/topic`, or whatever was set as the topic, is the actual displayed topic? This is already the case so :clap: on this: HTML tags must be kept and not removed from the topic.

- Oh My. Multiline:

  screen shot 2015-09-26 at 17 11 32

  Sadly we can't do any great magic at the moment. Simplest solution would be to `overflow: hidden;` it. Let's do that for now and open a ticket for long topics, would you be okay with that?

client/js/shout.js
astorije: @kalaksi I just tested `/topic I'm a "test"` and it works just fine. Probably best if you could amend your commit and remove the parenthesis (but keep the comment on `.text()`, it's useful and relevant here).

src/plugins/irc-events/topic.js
astorije: Hey @kalaksi, really sorry if you took it that way, I wasn't trying to be mean or anything, just humorous. I would never do that to someone who nicely contributed! :-)
So yes, I am always in favor of improving the code. Would be great if you can apply my suggestion! Thanks for spotting that.

kalaksi: Oh, it's fine :). Sometimes it's easy to make a mistake when interpreting the tone. But yeah, I'll do the changes later.

JocelynDelalande:
> Oh My. Multiline:
> [...]
> Sadly we can't do any great magic at the moment. Simplest solution would be to overflow: hidden; it. Let's do that for now and open a ticket for long topics, would you be okay with that?

That's more bad than it seems. Topics are quite often long on IRC channels and mention chan rules or things you should read before asking a question. So, it's quite anti-internet-social-practices to just hide it.

IMHO, a minimum would be to provide some kind of full topic text on hover (tooltip), (no precise idea of how to do it, would it work with  a `title` attribute ?).

JocelynDelalande:
@kalaksi thanks for that contrib :). Reviewed and tested on my side, pretty ok, with the two same remarks as @astorije : 

- multi-line is clunky (but see my other comment, on @astorije  suggestion), 
- some strange capitalization happens

kalaksi:
I did the improvements as requested and started looking into the CSS to make long topics work. I noticed that it has "text-transform: capitalize" so that's where the capitalization seems to be happening. IMO, it's unnecessary and may be confusing so I'll remove it, too.

kalaksi:
I found that there was this .header with "!important" which imposed a 50px line-height on the topic. Thinking it'd be simpler to just test it than to find every place that it affects, so I tested the UI with chromium and firefox and couldn't see anything that removing the "!important" changed  (except made long topics work :P) so I pushed the changes as such. What could use 50px line-height anyways..?

JocelynDelalande:
@kalaksi you're so reactive :)

> I found that there was this .header with "!important" which imposed a 50px line-height on the topic. Thinking it'd be simpler to just test it than to find every place that it affects, so I tested the UI with chromium and firefox and couldn't see anything that removing the "!important" changed (except made long topics work :P) so I pushed the changes as such. What could use 50px line-height anyways..?

Hey, that's cool you started working on something more long-term than my tooltip suggestion :)

From what I see, it broke nothing new, however, and in fact, it's getting better on large screen, as long as the topic is 2 lines or less. However, it's still broken on small devices : 

capture d ecran 2015-09-26 a 23 49 39

astorije:
Alright, my turn :-) Thanks for answering so quickly @kalaksi!

So it appears solving long topics will not be a 9-char easy fix, but thanks for trying. The current solution as @JocelynDelalande posted is not acceptable as it will make Shout on mobile most of the time unusable (anytime topics are long, which is most often).

@kalaksi, fact is we should never be using `!important`, but my experience showed that removing an existing one always breaks stuff (which is one of the reasons why we should never be using it...). In our case, I wouldn't be surprised if other themes don't take this change too well.

Jocelyn, I agree that hiding the end of a topic is not great, but it's always **better** than the current solution: no topic at all! Therefore, here are my comments to have this PR quickly acceptable:
- Revert the `!important` removal (:cry:)
- `overflow: hide;` the topic
- Set the topic in a `title` attribute as @JocelynDelalande suggested, good idea for a short-term solution
- **Open an issue to handle long topics.** Could be a single line with a horizontal scrollbar (xchat, IIRC), could be up to 2 lines with a vertical scrollbar, could be a single line and a "Show more" link that expands, ... Whatever, not the issue here.

Obviously, everything lies in the last step. Fact is this is not easy and requires discussion, while this PR is mostly ready. Let's get this simple addition sorted while we think of solving the bigger issue.

kalaksi:
Yeah, I tested it too quickly. I looked at the CSS again and IMO the whole CSS would need a larger overhaul. For one, it's messy. Also, too many things have `display: absolute` (and `!important`, `line-height` is also often used for some reason), so I couldn't get both the header (topic etc.) to grow dynamically at the same time with the chat message block without making larger changes. Also, would be nice to avoid defining stuff in pixels and make the leave-button and menu-button part of the right-side panel instead of the header-bar.

Removing the influence of the `line-height` apparently changed the position of the topic a bit. AFAIK  `#window .header` is only used for the header-bar that holds topic and icons but doesn't have other elements, so I'd rather leave out the `!important` and just align those few elements by removing separate top margins and adding paddding to the whole header. `overflow-y: auto` could be better for long topics instead of a title. What do you think?

astorije:
> I looked at the CSS again and IMO the whole CSS would need a larger overhaul. For one, it's messy. [...]

Yep! I know @erming is considering a refactor of the UI and this would be one aspect. We'll see how we prepare the (currently empty) [roadmap](https://github.com/erming/shout-roadmap) for this.

> make the leave-button and menu-button part of the right-side panel instead of the header-bar.

Not sure about the whole UI, but one sure thing is that we need mockups, discussions, consensus, ... Therefore, don't make those changes as part of this PR while your PR is mostly ready :-)

> AFAIK  `#window .header` is only used for the header-bar that holds topic and icons but doesn't have other elements, so I'd rather leave out the `!important` and just align those few elements by removing separate top margins and adding paddding to the whole header.

I would prefer you don't touch this ugly duckling. Chances are you are breaking another theme, or something else... I'd rather avoid potentially breaking something while you add a nice feature. Let's keep the removal of `!important` as part of the UI refactor or a later PR if refactoring doesn't happen soon.

> `overflow-y: auto` could be better for long topics instead of a title. What do you think?

Good point but, actually, I'd rather have an `overflow-x: auto;` instead: it makes more sense visually (scrolling vertically on a single line is rather ugly) and it seems much better on mobile (on a vertical mobile we would have as many lines as words, almost).
However, whenever using an `overflow`, you'll have some work as it doesn't apply on `span` directly.

Let me know if you have any other concern.
I can't wait to see this upstream! \o/

kalaksi:
Yea, I'm not going to implement any bigger changes :-). If it could have been done in a few changes that wouldn't have affected many elements, then I think it would have fitted in the scope of this PR. I was just saying what was on my mind. 
Btw, instead of moving leave-button etc. to right-side panel, another solution would be to have them inside full-height block so topic wouldn't overflow underside them and so on. That was the point behind the idea.

I think you're right about overflow-x, so I'll do it like that. That way we can also keep the `!import` and still have a sensible scrolling with the topic!

JocelynDelalande:
Just tested it :+1:

Good job for discussing a pragmatic and consistent solution for now guys <3

astorije:
Looks good to me too!
One last thing @kalaksi, could you squash your last 4 commits altogether please (using the description for details as you did in some)? After that I'll be happy to merge this :-)

src/plugins/irc-events/topic.js
```diff
@@ -9,27 +9,24 @@ module.exports = function(irc, network) {
...
-		}
 		var topic = data.topic;
+
 		var msg = new Msg({
 			type: Msg.Type.TOPIC,
 			mode: chan.getMode(from),
 			from: from,
 			text: topic,
-			self: self
+		        self: (from.toLowerCase() === irc.me.toLowerCase())
```
astorije: Argh, I missed this part, sorry! Can you replace the 4 spaces with a tab, please? :-(

kalaksi:
Bleh, damn tabs... Expanded the commit message, too, since it was left a bit terse after squashing

src/plugins/irc-events/topic.js
kalaksi: Bleh, damn tabs... Expanded the commit message, too, since it was left a bit terse after squashing

astorije:
Thanks @kalaksi and congrats on the good work!!

PR #418, opened 2015-05-04
client/js/shout.js: tag Notifications to reduce notification spam

williamboman:


astorije:
Some reads for the curious:

- https://developer.mozilla.org/en-US/docs/Web/API/Notification/tag
- https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API#Replacing_existing_notifications

@williamboman, this creates one "notification area" per channel (so N messages on 1 channel trigger 1 notification, but 1 message on N channels trigger N notifications). Is that what you were aiming at? If so, you get my :+1:.

astorije:
@williamboman, I forgot to follow-up on this, sorry. This is a very simple fix but has some nice impact, so it would be great to have your comments on my question :-)

I'm flagging this for second review as I believe this can be merged either way.

williamboman:
> Is that what you were aiming at?

Yep!

astorije:
Perfect, thanks! Full :+1: :-)

xPaw:
:+1:

PR #449, opened 2015-07-23
Prevent browsers leaking referrer headers

vtduncan:
This prevents browsers sending Referrer headers when the user clicks a link or views an image from within Shout. I think this is a safer default because users might not expect to reveal their domain simply by following a link.

The meta tag is part of the [Referrer Policy spec](https://w3c.github.io/webappsec/specs/referrer-policy/) and already works in Chrome and Firefox. (Chrome currently has a bug where this only works for following links, not embedding images. Hopefully it gets fixed soon.)

Thanks for working on Shout!

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -8,6 +8,7 @@
 	<meta name="apple-mobile-web-app-capable" content="yes">
 	<meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
 	<meta name="mobile-web-app-capable" content="yes">
+	<meta name="referrer" content="no-referrer">
 
 	<title>Shout</title>
 


richrd:
This would be great! :+1:

derhuerst:
There is an additional technique. It is somewhat ugly, but it definitely is the most effective way. It is [supported on every browser (IE>=10)](http://caniuse.com/#feat=datauri).

**Just encode links as Data-URI with inline HTMl that redirects to the target URL.**

This also has some disadvantages:

- As far as I remember, this shows up as another entry in the browsing history.
- I think most of the browsers' prefetching mechanisms (DNS resolution, malware checking, caching) will fail.

vtduncan:
Hi @derhuerst, thanks for the suggestion!

I did a bit of testing around the different techniques for restricting Referrer headers. Here is a [test suite](https://referrer.vtllf.org/) for comparing them, including Data-URIs. I'd love to know what works in IE if you have access to it and get the chance!

I went with the `<meta name="referrer">` tag in this case because:

- It only needs to be added once per page, no chance of missing individual links in the future.
- It doesn't interfere with browser behaviors like copying/pasting links and previewing the URL on mouse over.
- Also prevents sending Referrers for Shout's image previews.
- Because Shout uses Google Fonts, Referrers can also get sent there automatically. The meta tag fixes that too.

There are two other methods that I would also support fully: Content-Security-Policy either as an HTTP header or meta tag. Both of these have the same benefits and also work today *even in Chrome*. I think the `<meta name="referrer">` is simpler and more direct because it only has one purpose, but any of these fixes would be great.

vtduncan:
(On second thought, previewing link URLs isn't really a problem for IRC. It's all plain text duh!)

derhuerst:
> I'd love to know what works in IE if you have access to it and get the chance!

Unfortunately, I do not. However, for those really eager enough, there are [VMs by Microsoft as part of the modern.ie program](http://dev.modern.ie/tools/vms/mac/).

> Because Shout uses Google Fonts, Referrers can also get sent there automatically. The meta tag fixes that too.

This is interesting, maybe I'll open up another issue: I'd love to see those fonts getting served by shout (maybe as an option, if there not enough options yet :P), because I want to host shout in my local network and be as independent from a working internet connection as possible.

tobmaster:
Merge +1 :)

JocelynDelalande:
:+1:, good for privacy.

astorije:
:+1: 

As of when I am writing this, the spec (http://www.w3.org/TR/referrer-policy/) is only a FPWD, so much can be changed (although the editor's draft is only a week old, so plenty has been done since August 2014!).
Merging but we need to make sure the attribute doesn't change over time.

PR #386, opened 2015-03-29
Added Themes Selector

AmShaegar13:
Addresses #90. Added themes selector to config page. Themes are read from client/themes directory. Selected theme is saved in settings cookie.

PR #454, opened 2015-08-05
Add theme selector to settings

### LINKED ISSUE #452: CSS themes per user
It'd be a nice thing to have a list of CSS themes to pick from and remember it per-user.

floogulinc:
Adds a theme selector to the settings that allows users to choose a different theme. Themes are managed in config.js and consist of only a CSS file. The default config.js now comes with the default theme and the other 3 that are come in the themes folder preconfigured. 

After updating, admins should either add the new theme stuff and remove the old variable, or regenerate a new config.js.

Closes #452.

screen shot 2015-08-05 at 04 25 08

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -13,7 +13,7 @@
 
 	<link rel="stylesheet" href="css/bootstrap.css">
 	<link rel="stylesheet" href="css/style.css">
-	<link id="theme" rel="stylesheet" href="<%= theme %>">
+	<link id="theme" rel="stylesheet">
 
 	<link rel="shortcut icon" href="/img/favicon.png">
 	<link rel="icon" sizes="192x192" href="/img/touch-icon-192x192.png">
@@ -218,6 +218,20 @@
 								<label class="opt">
 									<input type="checkbox" name="colors">
 									Enable colored nicknames
+								</label>
+							</div>
+							<div class="col-sm-12">
+								<label class="opt">
+									<select id="themeselect" name="themeselect">
+									<% themes.forEach(function(theme) { %>
+										<option value="<%= theme.path %>" 
+										<% if (theme.default) { %>
+										selected
+										<% } %>
+										><%= theme.name %></option> 
+									<% }) %>
+									</select>
+									&nbsp;Theme
 								</label>
 							</div>
 							<% if (typeof prefetch === "undefined" || prefetch !== false) { %>
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -346,6 +346,17 @@ $(function() {
 		}
 	}
 
+	if($.cookie("theme") != null) {
+		$('#themeselect').val($.cookie("theme")).change();
+	}
+
+	$('#theme').attr('href',$('#themeselect').val());
+
+	$('#themeselect').on('change', function() {
+		$('#theme').attr('href',$(this).val());
+		$.cookie("theme", $(this).val()), options, {expires: expire(365)};
+	});
+
 	settings.on("change", "input", function() {
 		var self = $(this);
 		var name = self.attr("name");
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -35,12 +35,57 @@ module.exports = {
 	bind: undefined,
 
 	//
-	// Set the default theme.
+	// Themes
 	//
-	// @type     string
-	// @default  "themes/example.css"
+	// @type     array
 	//
-	theme: "themes/example.css",
+	themes: [
+		//
+		// A theme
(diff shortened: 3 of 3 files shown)


richrd:
:+1: I would love this! Although I feel that the theme list should be automatically detected from the theme folder. I like to create my own themes or modify existing ones after copying them to a different file. I understand this requires extra hassle with client-server communication, but I think detection would be optimal solution and offers the best user experience.

That said, I still feel this could be merged any time :)

floogulinc:
Well, adding themes with that only requires adding a few lines to the config file for your instance, It also allows for the theme file to be at any path, or even a URL to one hosted elsewhere.

richrd:
Ok, looks good. For my requirements I just need config reloading after this is implemented :+1:

heubergen:
FYI: Works for me, I've updated my files manually.

client/index.html
```diff
@@ -214,12 +214,26 @@ <h1 class="title">Settings</h1>
 							<div class="col-sm-12">
 								<h2>Visual Aids</h2>
```
astorije: I would prefer a new "Themes" category rather than adding this to "Visual Aids".

client/index.html
```diff
@@ -214,12 +214,26 @@ <h1 class="title">Settings</h1>
...
 									Enable colored nicknames
 								</label>
 							</div>
+							<div class="col-sm-6">
+								<label class="opt">
+									<select id="themeselect" name="themeselect">
+									<% themes.forEach(function(theme) { %>
+										<option value="<%= theme.path %>" 
+										<% if (theme.default) { %>
+										selected
```
astorije: What about `default` not being a property on the theme (but maybe a different one outside of `themes`, like `defaultTheme`), so that we avoid situations with multiple defaults?

client/js/shout.js
```diff
@@ -346,6 +346,17 @@ $(function() {
 		}
 	}
 
+	if($.cookie("theme") != null) {
+		$('#themeselect').val($.cookie("theme")).change();
+	}
+
+	$('#theme').attr('href',$('#themeselect').val());
```
astorije: Space missing: `'href', $('#themeselect')`

client/js/shout.js
```diff
@@ -346,6 +346,17 @@ $(function() {
...
+
+	$('#themeselect').on('change', function() {
+		$('#theme').attr('href',$(this).val());
```
astorije: Missing whitespace here as well.

defaults/config.js
```diff
@@ -35,12 +35,57 @@ module.exports = {
...
+		// @type     object
+		//
+		{
+			//
+			// Name
+			//
+			// @type     string
+			// @default  "Default"
+			//
+      		"name": "Default",
```
astorije: Whitespace instead of tab here and in lines below.

astorije:
Hey @floogulinc, thanks! I left a few comments inline, but I think we are very close to a shippable solution overall, great! :-)

floogulinc:
@astorije with all the whitespace stuff, I was just following what was already used in the rest of the file.

client/index.html
floogulinc: @astorije Note that in Shuo we renamed this area "Visuals"

astorije: Thanks for the precision! I would use the word "Theme" though, or "Interface" maybe, but I prefer "Theme". And we'll include there everything related to design/look options.
What do you think?

astorije:
Hey @floogulinc, actually, most of the code is `func(arg1, arg2)` and not `func(arg1,arg2)` so let's keep that version. :-)
Also, I might have been unclear in my last comment, but the problem is some lines of `defaults/config.js` are using spaces for indentation where they should be using tabs as the rest of the project.

I can't wait to see this released!! :-)

client/index.html
floogulinc: @astorije I honestly like calling it "Visuals" because it is specifically things relating to the visual appearance.

floogulinc:
@astorije OH, thats what you meant, I thought you were talking about the indenting.

defaults/config.js
floogulinc: @astorije I was just following what the rest of this file uses. It is all tabs, though that might have changed in the last 5 months.

astorije:
Argh, it's a shame, this is so close to a shippable solution :-(

astorije:
Noticed a potential duplicate: https://github.com/erming/shout/pull/386
I'll review and check if they are indeed duplicates or not.

PR #455, opened 2015-08-16
Fix XSS vulnerability

ohaal:
Quick fix for a nasty little XSS vulnerability, see private message for PoC.


astorije:
Please merge this, this is a nasty one! :-)

xss

floogulinc:
Please, run `grunt` and commit it's changes to the template js. Otherwise, the only way this fix works would be for users to run it themselves.

ohaal:
Thanks, I didn't think about `grunt` not being part of the install process.

PR #474, opened 2015-09-23
Replace menu icon with users icon for userlist toggle and smooth the 2 icons

astorije:
On mobile, the left icon for the main menu was the same than for the userlist. A 3-bar icon for a userlist 
I have changed it to a "users" icon, and applied font smoothing otherwise it looks disastrous.
The result:

Userlist icon

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -117,10 +117,17 @@ button {
 	margin: 6px 12px 0 -12px;
 	width: 36px;
 }
-#viewport .lt:before,
-#viewport .rt:before {
+#viewport .lt:before {
 	font: 14px FontAwesome;
 	content: "\f0c9";
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
+}
+#viewport .rt:before {
+	font: 14px FontAwesome;
+	content: "\f0c0";
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
 }
 #viewport .rt {
 	display: block;


astorije:
@erming, @JocelynDelalande, may I have your :+1: on this?

erming:
I don't like that icon.

Maybe we should just make it a button instead, like this:

button_hide_users

Makes it more obvious what it's supposed to do.

astorije:
Mmh you don't want that for mobile versions! Especially since we'll soon be solving the long topic issue (see #409).

I know you have plans for a better front-end with a refactor, this PR is just an easy quick fix to remove the current two 3-bar menus that are very confusing. Could we go for this until we discuss UI more deeply in our roadmap? :-)

FYI there is no other icon in Font Awesome that would do here, I checked them *all*...

JocelynDelalande:
@astorije thanks, that had to be fixed. 

The 3-bar icon used curently is clearly not the right ui element here.

I would be slightly more in favor of a button, for consistency and explicitness. It's not obvious this icon is clickable.

@astorije @erming What would you think about that icon on a button ?

> FYI there is no other icon in Font Awesome that would do here, I checked them all...

This guy may be dangerous, he's having strange readings ;-)

astorije:
> I would be slightly more in favor of a button, for consistency and explicitness. It's not obvious this icon is clickable.
> 
> @astorije @erming What would you think about that icon on a button ?

The problem is that it takes extra space, which is undesired on mobile, and with topic coming in (see #409), it's gonna be a problem. I was actually thinking of replacing the `Leave` button with [that icon](https://fortawesome.github.io/Font-Awesome/icon/sign-out/) to help usage on mobile.

---

Here is a bigger picture I notice after parsing loads of UI-related PRs and issues: when it comes to UI, a lot of things that are suggested, submitted, chosen as solution, are quick fixes because of some other reason. Add to that that @erming is aiming for a refactor, I do think that this the path we are looking at: quick and easy fixes for the short term while recording everything we want in our [roadmap](https://github.com/erming/shout-roadmap), including mockups, before starting something nice. We'll need help from UI/UX smarty pants, such as @bnvk :-)

What this means here is that I'm okay with non-ideal icons until we solve something more globally: buttons in the header that are identifiable as buttons and unobtrusive for the mobile user.

richrd:
I think the icon solution is elegant and good since it doesn't waste space. :+1:

The icons should have tooltips to indicate the action they perform (on desktop). It would also be optimal if the markup contained both an icon and the text, so the theme should decide how to show the button (or the icon). The choice could also be made based on the screen size. 

PS @reming I really love your project, so sorry that I don't agree with you on this one ;). Anyway I think the user (and theme developers) should have more choice on these things. Just how I feel :beers:

astorije:
Ping! Can we have this as a short term solution until we find better? It's still always better than the current situation... @JocelynDelalande?

JocelynDelalande:
@astorije The inconsistency of this part of UI makes me cry, but this PR do not introduce it after all, so I give my :+1: to this icon-by-icon replacement.

astorije:
What do you mean, @JocelynDelalande?

JocelynDelalande:
> What do you mean, @JocelynDelalande?

an action-button (leave) aside an icon-button (users list) aside a label-icon (search)

astorije:
Agreed, an all-icon, before some sort of contextual menu, will please the mobile users. :-)
Will send a PR!

PR #481, opened 2015-09-27
client/js/shout.js: prevent default action when clearing chat

### LINKED ISSUE #453: Chat clear keybinding conflict
Modern browsers offer the keybinding `Ctrl + L` to select the location bar.
Shout implements this shortcut to clear the chat history since #165.

Maybe this shortcut should be changed or the browser action should be prevented.

williamboman:
Closes #453.

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -723,11 +723,11 @@ $(function() {
 
 	Mousetrap.bind([
 		"command+k",
-		"ctrl+l",
 		"ctrl+shift+l"
 	], function (e) {
 		if(e.target === input[0]) {
 			clear();
+			e.preventDefault();
 		}
 	});
 


williamboman:
I think this is a suitable solution. Shout is an application which most users constantly keep opened/pinned in the background<sup>1</sup>, so overriding native browser keybindings is not that big of an issue, IMO.

--

<sup>1 - not subjective at all, nope</sup>

JocelynDelalande:
> I think this is a suitable solution. Shout is an application which most users constantly keep opened/pinned in the background1, so overriding native browser keybindings is not that big of an issue, IMO.

I disagree with breaking that default browser keybinding, especially for people using their browser exclusively/mostly via keyboard.

(and my opinion is the one of someone using Ctrl+L all the day in terminals ;-)

Maybe `Ctrl+shift+L` ? Is this even feasible on a browser ?

What do others think ?

astorije:
I also believe we should not be overriding most common shortcuts. On Windows/Linux (as far as I know), <kbd>Ctrl</kbd>+<kbd>L</kbd> moves focus to the location bar, which is a pretty big deal.

But fact is, it's pretty difficult to get "safe" shortcuts:
- https://stackoverflow.com/questions/3329420/what-are-cross-browser-cross-os-safe-keyboard-shortcuts-usable-for-web-applica
- http://www.w3.org/TR/html401/interact/forms.html#adef-accesskey
- http://lea.verou.me/2011/12/on-web-apps-and-their-keyboard-shortcuts/
- https://medium.com/@sashika/j-k-or-how-to-choose-keyboard-shortcuts-for-web-applications-a7c3b7b408ee

JocelynDelalande:
@astorije interesting links , thanks for sharing :)

Back on earth after reading that, I think <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>L</kbd> is still a good compromise between not breaking any well known browser/window manager integration and using something that power-users can memorize easily, because it's close to <kbd>Ctrl</kbd>+<kbd>L</kbd>.

(maybe I wrote this comment just because I wanted to try the nifty  `<kbd>` tag @astorije used in his comment ;-)

williamboman:
I've removed <kbd>Ctrl</kbd>+<kbd>L</kbd>, but kept `.preventDefault()`.

JocelynDelalande:
:+1:

richrd:
:+1:

astorije:
:+1:

astorije:
@williamboman, that would be great if you could update the doc accordingly: http://shout-irc.com/docs/client/keyboard_shortcuts.html#clear-buffer

The corresponding file is at: https://github.com/erming/shout-website/blob/gh-pages/_docs/client/keyboard_shortcuts.md#clear-buffer

PR #489, opened 2015-09-28
Fix the link plugin test

astorije:
> Travis should work now

Taken from https://github.com/Shuo-IRC/Shuo/pull/31.

PR #493, opened 2015-09-28
Fix tests

williamboman:


test/plugins/link.js
```diff
@@ -1,4 +1,7 @@
 var assert = require('assert');
+var path = require('path');
+
+process.env.HOME = path.join(process.cwd(), 'test', 'fixtures');
```
williamboman: eh, it does the job though

astorije:
Well, I was working on that right now as you can see on #489... but sure, we can use your version. Let me give up what I was doing and review yours :-)

package.json
```diff
@@ -39,6 +39,7 @@
     "socket.io": "~1.0.6"
   },
   "devDependencies": {
+    "extend": "^3.0.0",
```
astorije: Mmmh no thanks, a new dependency to fix one test, knowing that the reason itself should be fixed itself ideally (i.e. remove the config read in the link plugin...), let's not go towards that path.

test/fixtures/.shout/config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = require('extend')(require('../../../defaults/config.js'), {
+	prefetch: true
+});
```
astorije: AFAIK, this works too and removes the need for the dependency:

```
var config = require('../../../defaults/config');
config.prefetch = true;
module.exports = config;
```

williamboman: True, fixed.

package.json
williamboman: Removed

test/fixtures/.shout/config.js
```diff
@@ -0,0 +1,4 @@
+var config = module.exports = require('../../../defaults/config.js');
```
astorije: Mmmmh, please don't :-)
My snippet was not much longer and more readable IMO.

williamboman: This is pretty idiomatic in node-land though

test/fixtures/.shout/config.js
williamboman: Note that this will modify the cached `defaults/config.js` object (which is why extending is might be a good idea).

test/fixtures/.shout/config.js
astorije: Yeah, I have seen that on occasions (like, two), but it doesn't mean it's not ugly. I'd really go for something more explicit instead, which is how it's being used most of the time.

test/fixtures/.shout/config.js
astorije: Indeed, but I don't expect this config loading to stay there before other tests are added, and we won't have to do that config rain dance to satisfy the test.

astorije:
It's a bit hackish, but so is the reason why we need this, and having the test pass is more pressing than refactoring that bit of the code.

You have my :+1:, awaiting for second reviewer.

JocelynDelalande:
did quick review, :+1: for me as the result is here : test<strike>s</strike> <strike>are</strike> passing ! Thanks @williamboman 

@williamboman just see my inline comment, and fix it if relevant, otherwise, :+1:

test/plugins/link.js
```diff
@@ -26,7 +29,7 @@ describe('Link plugin', function() {
     });
 
     this.irc.createMessage({
-      message: 'http://localhost:9002/basic'
+      message: 'http://127.0.0.1:9002/basic'
```
JocelynDelalande: is this actually fixing things ? if not, please revert this change, cause it enforces IPv4 over IPv6.

astorije: @JocelynDelalande, I do think Travis CI cannot pass without that, as proven by https://github.com/Shuo-IRC/Shuo/pull/31, but I'm not entirely sure why.

williamboman: https://github.com/erming/shout/blob/master/src/plugins/irc-events/link.js#L21 makes it look it ignores `https?://localhost` files

JocelynDelalande: @williamboman Wow, weird.

astorije: The reason is that we do not want to prefetch links that point to `localhost`. But that's actually a flaw: if we don't want that, we should also prevent `127.0.0.1`, `[::1]`, ... So here again, the problem should be solved at a deeper level :(

JocelynDelalande: > The reason is that we do not want to prefetch links that point to localhost

why ?

astorije: I don't know, but that's the intent :-)

floogulinc:
BTW, i noticed this:
>This job ran on our legacy infrastructure. Please read [our docs on how to upgrade](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade)

This is probably a cause of our Travis build problems and I bet #489 would work fine if it was upgraded.

JocelynDelalande:
@floogulinc no, it's because of hardcoded exception to prefetch, see #496

test/plugins/link.js
astorije: @williamboman, could you amend and remove that from second commit so that we keep `localhost`? [This commit](https://github.com/erming/shout/commit/58aa236dcd6df1011a239568201da41b893e9f84) now makes this unnecessary.

williamboman: Fixed, added a `test.js` file for bootstrapping tests.

astorije:
Argh, I wish you had shared that design before spending your time on it :-(

I'm sorry but I'm :-1: on the `test.js` file. At the moment, the only benefit is to be able to specify `node test` in the `package.json` rather than `mocha test/**/*.js`. I don't see it as a benefit, because devs and CIs will run `npm test` anyway. The drawbacks here are a new dependency, a new 20-line file and the need to know mocha's dev API whenever we need to change the runner. On top of that, it adds a new level of indirection to know what's behind `npm test`.
After re-reading, sorry, I missed one other benefit: centralizing the fixtures.

Considering we only have one test at the moment, this is truly overkill.
Here is my proposal: please switch back to the "simple" version we discussed earlier, as @JocelynDelalande and I were both OK with that. Meanwhile, open a PR with your added changes. It's likely it will be closed until relevant, but at least your work will not go to waste.
Are you OK with that?

williamboman:
>At the moment, the only benefit is to be able to specify node test in the package.json rather than mocha test/**/*.js

The benefit is having an entrypoint for running and most importantly bootstrapping tests, it has nothing to do how the scripts look like in `package.json`, that's completely irrelevant. Just specifying a glob of files for `mocha` to test in your `package.json` is fine in small modules where the only thing you're testing is the module itself.

In a project like this, with many different parts (that *should* be unit tested) `mocha test/**/*.js` is a bit primitive. `test.js` could be abstracted into a grunt task instead, if that would make things clearer.

>The drawbacks here are a new dependency, a new 20-line file 

I don't see why you're so afraid of dependencies, those are backbone of the entire Node.js ecosystem. We have `grunt` as a dependency, which in turn depends on `glob`, so it's already there. It is not de-duped since the versions are mismatching, I could downgrade the dependency to be compatible with `grunt`'s `glob` dependency - `npm` will then handle de-duping these dependencies - reserving the exact same file size of `node_modules/`.

A new 20-line file that expresses how tests are executed, and its bootstrapping process, rather than a very ambiguous `process.env.HOME = ...` in the midst of a unit test file.

williamboman:
Updated with a cleaner solution (for now).

astorije:
> Updated with a cleaner solution (for now).

Now we're talking :-)
I :+1: this version and will answer your longer comment when I am not about to board a plane.

@JocelynDelalande, still OK with this PR? If so, feel free to merge.

williamboman:
>I :+1: this version and will answer your longer comment when I am not about to board a plane.

No need, I understand and agree with your view (it really was out of scope for this PR as well).

JocelynDelalande:
@williamboman thanks a lot !

small is beautiful :)

PR #494, opened 2015-09-29
Migrate from legacy to container-based infrastructure on Travis

floogulinc:
http://docs.travis-ci.com/user/migrating-from-legacy/

Should make PR #489 work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,4 @@ node_js:
 before_script:
   - mkdir ~/.shout
   - cp defaults/config.js ~/.shout
+sudo: false


astorije:
I don't think this will fix the test, but it's generally a good idea to migrate to Travis CI's new infrastructure, so I :+1: this PR no matter what.
Not affecting my review, but could you rebase this branch so that we know if it fixes or not?

Awaiting for second review before merge.

floogulinc:
@astorije oh yeah, I forgot that on Shout prefetch is now disabled by default.

JocelynDelalande:
:+1:

Issue #388, opened 2015-03-30
Don't fetch `localhost` links


jsocol:
When someone posts a `localhost` link, there's still a fetch, which is both obviously not valuable (since it seems to fetch Shout) and potentially risky (I don't know of any immediate exploit but I haven't looked either).

Planning to submit a patch, just wanted to write it down for myself (or someone who beats me to it).

erming:
I don't think there's any potentinal risks here. It can only fetch public files (not internal) hosted from http://localhost/. The server would have to serve the file you link to, otherwise it will just 404. Also, it's capped to 10mb max-size for a document it tries to fetch.

You've probably already seen the code, but I link it anyway:
https://github.com/erming/shout/blob/master/src/plugins/irc-events/link.js

I agree with you though, it's not valuable and it should be blocked.

erming:
Fixed here: https://github.com/erming/shout/commit/29b66ff0ec6bdd67c7c3168c383efe00bf94f75f

The link is still clickable but it will not be fetched and have its headers read etc.

Thanks man!

jsocol:
Thanks! I got busy and didn't have a chance to follow-through.

PR #496, opened 2015-09-29
Remove prefetching exception for localhost

JocelynDelalande:
This reverts commit 29b66ff0ec6bdd67c7c3168c383efe00bf94f75f.

Rationale:

1) It's not a security feature (abuse of prefetch can be on any server it's not
   more/less risky on localhost), it's pseudo-security measure
2) It's not to us to judge if it has no use-case (in fact it has, ex: two dev
   speaking and experimenting about urls of their local site/app instance,
   local web apps...)

refs #388

@erming  @astorije waiting for your +1 to merge.

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -18,9 +18,6 @@ module.exports = function(irc, network) {
 		var links = [];
 		var split = data.message.split(" ");
 		_.each(split, function(w) {
-			if (w.match(/^(http|https):\/\/localhost/g)) {
-				return;
-			}
 			var match = w.indexOf("http://") === 0 || w.indexOf("https://") === 0;
 			if (match) {
 				links.push(w);


maxpoulin64:
> 1) It's not a security feature (abuse of prefetch can be on any server it's not
more/less risky on localhost), it's pseudo-security measure

Actually, yes it is. Well, it should, but right now it's merely a bandaid over the issue as it doesn't even check for 127.xx.xx.xx. Services listening on localhost should only be used by localhost. Often those would include backend servers running behind (and potentially password protected by) an nginx. Granted, those services should have their own authentication, but many services are setted up to listen only on localhost because only local people are expected to access it. Running shout allows anyone to query that server, and it's not obvious for everyone that it does so.

I'd vote to make it a configurable blacklist that defaults to blocking all private IP ranges, and let users open it up if they want to (by removing the relevant blacklisted items). If running on a home computer, shout also exposes the entire private network for which one might have firewalled properly from the outside but not the inside (as the inside is *expected* to be able to access, lets say, intranet sites).

JocelynDelalande:
> I'd vote to make it a configurable blacklist that defaults to blocking all private IP ranges, and let users open it up if they want to (by removing the relevant blacklisted items). If running on a home computer, shout also exposes the entire private network for which one might have firewalled properly from the outside but not the inside (as the inside is expected to be able to access, lets say, intranet sites).

May be better like current filter, but IMHO, it's still illusion of security :

- still, the hostname(s) can be used to hit localhost or private networks
- you cannot know what is actually private network (think corporate networks with public addresses, think IPv6)

maxpoulin64:
> the hostname(s) can be used to hit localhost or private networks

It should resolve the name first before checking IPs, of course!

> you cannot know what is actually private network (t lhink corporate networks with public addresses, think IPv6)

True that, but I think it's relatively safe to assume that if you have a public IP, said IP is reachable from the outside. IPv6 networks usually have both public and private IPs, so if you listen to something private on the public IP you're kind of an idiot.


The idea is obviously not to handle all use cases, but at least deal with the most common ones, and warn users about it when they go to turn on prefetch.

JocelynDelalande:
> True that, but I think it's relatively safe to assume that if you have a public IP, said IP is reachable from the outside. IPv6 networks usually have both public and private IPs, so if you listen to something private on the public IP you're kind of an idiot.

Not at all, it means, that you use firewall for firewalling, not NAT (which is sane thinking).

My take on that: we can discuss and implement something like you suggest @maxpoulin64 , but, as our discussion shows, that's not a simple problem.

@maxpoulin64 some time to submit a PR on that ?

At the moment, I'd prefer that we get back to clean room: having nothing is better than having something pretending to be a protection but being an illusion of protection.

JocelynDelalande:
@erming @astorije may I have your :+1:  ?

astorije:
I fully agree with @JocelynDelalande: this is an attempt to secure the `localhost` while we only lock the front door. Even though we wanted to prevent fetching from localhost, who are we to decide if it should or should not be blocked?
Situation you described is not a problem @maxpoulin64: you are not "accessing" the private network, you are just asking shout to load and display it to the other user, you're not entering it.

That being said, I agree that it would be nice to allow blacklisting at server configuration (since prefetching is done on the server).
And actually, that leads to my other worry: prefetching should be done on the client in my opinion. What do you think about that?

Regarding that PR though, it's a :+1: and a merge!

floogulinc:
@astorije Prefetching on the client would lead to issues with the ability to obtain user's IPs and locations just by posting a link in chat, but doing it on the server also has issues...

JocelynDelalande:
> And actually, that leads to my other worry: prefetching should be done on the client in my opinion. What do you think about that?

Well, I don't know browsers enough, but wouldn't that allows to trigger GET URLs on websites you are logged-on ?

With badly designed web services/sites (GET requests doing write actions), it can be automatic [CSRF attack](https://en.wikipedia.org/wiki/Cross-site_request_forgery), huray :)

JocelynDelalande:
The more I think about it, the more I think the decent way to implement it is in a plugin that prefetches over [TOR network](https://en.wikipedia.org/wiki/Tor_%28anonymity_network%29).

astorije:
Yes, that's the sort of trade-offs we'll have to choose from. I am personally in favor of doing this on the client-side:
- in terms of usage, it reflects more what the client sees, if the given URL, for example, reacts to a session. Another use case is the one Jocelyn gave: localhost will answer differently on the server and the client.
- in terms of "security" (not necessarily security issues here), doing it on the server-side gives information to a user about a server (say http://localhost/ on the server is an internal webpage, while on the client it's a dev project). Also, but that might be bad thinking, I'd rather have a client information compromised than a server information.

astorije:
@JocelynDelalande yes, we need client-side plugins for that, and then switch to TOR might be a good solution. Is there a reason why we should keep a non-TOR after that? I know that the place where I store my server forbids the installation of TOR relays, but maybe that's something completely different (/me doesn't know much about TOR).

JocelynDelalande:
> @JocelynDelalande yes, we need client-side plugins for that, and then switch to TOR might be a good solution. Is there a reason why we should keep a non-TOR after that? 

Let's discuss that later and in a dedicated ticket :)

> I know that the place where I store my server forbids the installation of TOR relays, but maybe that's something completely different (/me doesn't know much about TOR).

Nothing to deal with tor relays here it's being a bare tor client, which isn't likely to be a problem with your network operator/hosting place.

Nb: 

- **exit nodes** exposes operators/hosters  to speaking with the police at times (and that they sometime forbid) 
- **relays** are fine, but they can trigger heavy bandwidth usage
- **clients** are fine, there is no administrative/cost reason to forbid them (except if you want to spy your users…)

astorije:
> Let's discuss that later and in a dedicated ticket :)

Yes. After we discuss plugins :D

> exit nodes exposes operators/hosters to speaking with the police at times (and that they sometime forbid)
> relays are fine, but they can trigger heavy bandwidth usage
> clients are fine, there is no administrative/cost reason to forbid them (except if you want to spy your users…)

Right, the first 2 ones are forbidden on the network :-)
Thanks for the clarification.

floogulinc:
My point was, if prefetching was done on the client side and I wanted to obtain a user's IP address, I could send them a link and without them clicking it, it would make a request to my server and let me log it. This defeats hostname masking on the IRC network.

I'm more in favor of keeping how it is and blocking requests to localhost, possibly in a configurable way.

Xe:
> The more I think about it, the more I think the decent way to implement it is in a plugin that prefetches over TOR network.

That would mean adding a dependency though :)

richrd:
I feel that this should be configurable like this:

 1. None, don't prefetch (possibly default)
 > Fool proof. No way to exploit anything.

 2. Prefetch on server (possibly default)
 > Doesn't expose the client IP (which I think is really sensitive)

 3. Prefetch on client
 > Doesn't expose the server IP (which might be public anyway and the users know it)

**2 and 3 should have as much protection as possible against localhost etc links. But I understand that this is problematic.**

floogulinc:
@richrd I think this would be more complex than necessary, you would have to have the prefetch code in both the client and server.

JocelynDelalande:
> My point was, if prefetching was done on the client side and I wanted to obtain a user's IP address, I could send them a link and without them clicking it, it would make a request to my server and let me log it. This defeats hostname masking on the IRC network.

Fair point.

> I'm more in favor of keeping how it is and blocking requests to localhost, possibly in a configurable way.

Yep, but IMHO, that require what you said : configurable, and also another thing : taking it seriously with sane defaults and hostname resolution, otherwise, it's clearly pseudo-security.

astorije:
> My point was, if prefetching was done on the client side and [...]
> 
> I'm more in favor of keeping how it is and blocking requests to localhost, possibly in a configurable way.

I entirely agree, and if we continue work towards prefetching we need to ask ourselves these security questions, but keeping it on the server is a problem of feature/UX : it simply does not show what the client wants to see (if logged in, if based on geolocation, ...) !
So yes for security, but we must think of the good health of the feature itself and adapt our security discussions to that. Note that I will never say we should go for a security/privacy-faulty solution though :-)

maxpoulin64:
Guys, it's not even possible to prefetch on the client. Allowing one website to make requests to another site is *extremely* dangerous, as said website would be able to say, fetch your Gmail page and download all your mail, or approve pull requests on GitHub, or really anything. For that reason, browsers enforces the [same-origin policy](https://en.wikipedia.org/wiki/Same-origin_policy) and only allows requests to be done to the same domain that hosted the script, and blocks every other requests unless explicitely allowed by the other server ([CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing)), which is very rare and often restricted to a few select other external sites. Thus, there's no way to implement the prefetch feature on the client.

The only thing that could be possible on the client is to embed an image in an `` tag, which obviously works only with images, so we still need server assistance to fetch the page to know its content-type. At which point you're better just proxying everything through the server. The client is very limited in what it can do. The client should be considered as a dumb terminal.

JocelynDelalande:
> I entirely agree, and if we continue work towards prefetching we need to ask ourselves these security questions, but keeping it on the server is a problem of feature/UX : it simply does not show what the client wants to see (if logged in, if based on geolocation, ...) !

Yes, but preview is not core feature, moreover, prefetching from your browser, is doing something *on your behalf*, without your agreement, that's bad.

I have no problems with preview working only for public resources. + it's not to us to adapt to website geolocation-based hacks...

JocelynDelalande:
> Guys, it's not even possible to prefetch on the client. [...] Thus, there's no way to implement the prefetch feature on the client.

Yes, you're right about CORS, thanks for your clever input. So, that's just the biggest of many reasons not to fetch on client ;-)

astorije:
> For that reason, browsers enforces the same-origin policy

Fair point. I got lucky enough to work on a CORS-enabled project recently that I forgot that limitation. That also makes it impossible to make it a plugin on the client and, therefore, if we want to do it in a plugin on the server, we also need... a plugin mechanism there...
Although, not necessarily: to separate the code between this feature and Shout server itself (today it's mangled with IRC event receiving), the server could expose an entry point (`/prefetch`?) that the client, plugin or not, would dumbly gets its data from. Sort of an embed.ly alternative that we control ourselves, making sure we don't leak statistics or privacy.

> Yes, but preview is not core feature

Well that's an easy argument when someone disagrees ;-)

> + it's not to us to adapt to website geolocation-based hacks...

... Which are not necessarily hacks, such as load balancing or local (country) law-protected content (these are examples, please don't use them as authoritative cases to justify against)

PR #499, opened 2015-09-30
Replace Google hosted fonts with local ones

### LINKED ISSUE #497: Embed font instead of call google ones
Hello,

I noticed that the fonts seemed to be charged from https://fonts.googleapis.com when shout webpage is opened.

It should be better to embed them locally for better privacy.

floogulinc:
Closes #497


JocelynDelalande:
On the legal aspect, the fonts we embed are dual-licensed SIL open-font license / Apache.

Not an expert, but reading the SIL open-font license itself and the [associated FAQ](http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&item_id=OFL-FAQ_web#68092c0f) makes me to think that it's a MPL-like (file-level license) dedicated to fonts.

And for the record, an apache-licensed lib is [already included](https://github.com/erming/shout/blob/2cee0ea6ef5ee51de0190332f976934b55bbc8e4/client/js/libs/jquery/mousetrap.js)  within shout.

@floogulinc dare to add a note in the *License* section of the README mentioning the fonts licenses ?

astorije:
What @JocelynDelalande said :-)

JocelynDelalande:
:+1:

astorije:
:+1:

PR #503, opened 2015-09-30
Limit preview for large image files

olivierlambert:
Do not display a preview of content length greater than given in the config file (default: 512kb)

### DIFF
--- a/src/plugins/irc-events/link.js
+++ b/src/plugins/irc-events/link.js
@@ -52,6 +52,7 @@ module.exports = function(irc, network) {
 };
 
 function parse(msg, url, res, client) {
+	var config = Helper.getConfig();
 	var toggle = msg.toggle = {
 		id: msg.id,
 		type: "",
@@ -61,6 +62,9 @@ function parse(msg, url, res, client) {
 		link: url
 	};
 
+	if (!config.prefetchMaxImageSize) {
+		config.prefetchMaxImageSize = 512;
+	}
 	switch (res.type) {
 	case "text/html":
 		var $ = cheerio.load(res.text);
@@ -80,7 +84,12 @@ function parse(msg, url, res, client) {
 	case "image/gif":
 	case "image/jpg":
 	case "image/jpeg":
-		toggle.type = "image";
+		if (res.size < (config.prefetchMaxImageSize * 1024)) {
+			toggle.type = "image";
+		}
+		else {
+			return;
+		}
 		break;
 
 	default:
@@ -116,6 +125,7 @@ function fetch(url, cb) {
 			if (err) return;
 			var body;
 			var type;
+			var size = req.response.headers["content-length"];
 			try {
 				body = JSON.parse(data);
 			} catch (e) {
@@ -129,7 +139,8 @@ function fetch(url, cb) {
 			data = {
 				text: data,
 				body: body,
-				type: type
+				type: type,
+				size: size
 			};
 			cb(data);
 		}));
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -65,6 +65,17 @@ module.exports = {
 	prefetch: false,
 
 	//
+	// Prefetch URLs Image Preview size limit
+	//
+	// If prefetch is enabled, Shout will only display content under the maximum size.
+	// Default value is 512 (in kB)
+	//
+	// @type     int
+	// @default  512
+	//
+	prefetchMaxImageSize: 512,
+
+	//
 	// Display network
 	//
 	// If set to false Shout will not expose network settings in login


JocelynDelalande:
@olivierlambert please rebase on master (as tests have been fixed, and the one test we do have is somehow related to your PR ;-)). I'm reading/testing your PR right now.

JocelynDelalande:
@olivierlambert oh and on a more pendantic side, could you please squash your commits and reword commit message to explain what it does (it's good to mention the ticket but the commit msg should be sufficient to know what it's about).

JocelynDelalande:
First remark : why sending an empty preview ? better not to show any IMHO (the "…" thing on the first link pasted, which is a 50MB image).

screenshot-localhost 9000 2015-10-05 22-28-09

Second think, I'd suggest, setting the size on kB, there is little use setting it in octets IMHO :)

src/plugins/irc-events/link.js
```diff
@@ -129,10 +134,16 @@ function fetch(url, cb) {
 			} catch(e) {
 				type = {};
 			}
+			try {
+				size = req.response.headers['content-length'];
+			} catch(e) {
+				size = {};
+			}
```
JocelynDelalande: Maybe something more explicit than just catching all exceptions ?(I know, some existing code does like yo did), + are you sure `{}` is a valid size ? ;-)

JocelynDelalande:
@olivierlambert ok, reviewed, see inline comments for details.

Rather than fetching the image fully, a first timeserver-side and a second time, fully, client-side, I would suggest, doing a HEAD request to get the `content-size` header and decide, based on that, if the client should or not display a preview.

@olivierlambert what do you think about all my comments ?

olivierlambert:
* [ ] rebase on upstream to have tests running
* [ ] squashing commits
* [ ] better PR name
* [ ] content size in kB
* [ ] don't display any preview
* [ ] fix exception
* [ ] smart image fetching

olivierlambert:
@JocelynDelalande : I don't know where to start to avoid fetching the image as you requested. Would you like to do it in this PR or maybe as an optimisation in another one?

src/plugins/irc-events/link.js
```diff
@@ -60,7 +61,9 @@ function parse(msg, url, res, client) {
 		thumb: "",
 		link: url
 	};
-
```
JocelynDelalande: whiteline error

JocelynDelalande:
Ok, good for fixes :) thanks for your reactivity

> @JocelynDelalande : I don't know where to start to avoid fetching the image as you requested. 
> in link.js, for the case of the image I think you should replace 
> Would you like to do it in this PR or maybe as an optimisation in another one?

I was thinking on this one, but ok, that's acceptable to do it in a second time.

Starting point would be to call something issuing the HEAD request before [that fetch() call](https://github.com/olivierlambert/shout/blob/master/src/plugins/irc-events/link.js#L48).

JocelynDelalande:
@olivierlambert just tested again, you removed the empty "grey frame" but still the three dots on grey background are showing up on too big images.

Once that is fixed and the trivial whitespace thing, you'll have my :+1:

astorije:
@olivierlambert, any news on this? :)

JocelynDelalande:
@olivierlambert bump ? it's almost done :)

olivierlambert:
Didn't get any message/alert previously sorry :/

What about the "whiteline error"? I don't understand the issue there.

And IIRC (1 month ago, I could be wrong...), there is no trivial way to remove the three dots.

JocelynDelalande:
> What about the "whiteline error"? I don't understand the issue there.

You include in your commit the removal of a white line, which seems something unrelated to your edit.

> And IIRC (1 month ago, I could be wrong...), there is no trivial way to remove the three dots.

Do you think you could find some time to look again please ? (css may be the way ?) those three dots with no meaning at all are a really strange thing to see for the user.

olivierlambert:
1/ Oh okay, no problem with that.

2/ I'll try to take a look today

olivierlambert:
I can't avoid the `client.emit("toggle", toggle);` to happen, due to the current code architecture. AFAIK, there is two options:

1. create a ugly hack to re-hide the toggle somehow
1. find a better solution but this will lead to deep modifications

Or maybe I missed something, that's also another possibility ^^

olivierlambert:
Okay, you got exactly the same behavior for other type of content. Try to past a link to a PDF for example, you'll have the same issue.

It's not related to this PR, so I'll push my version without the white space, and I suggest to create a dedicated issue for the other problem :)

olivierlambert:
@JocelynDelalande PR done with the latest bits of master. Also tested it, that's fine.

JocelynDelalande:
> It's not related to this PR, so I'll push my version without the white space, and I suggest to create a dedicated issue for the other problem :)

Fair enough :) Thanks for digging it.

:+1:

astorije:
:+1: and merging.

Thanks a lot @olivierlambert!!! Hope to see you again soon!

olivierlambert:
You are welcome :)

PR #507, opened 2015-10-04
Fix Morning and Zenburn themes on mobile

floogulinc:
See Shuo-IRC/Shuo#46 for the original issue (fixed in Shuo-IRC/Shuo#85)

Before
--
The connect and settings buttons are invisible.

Morning | Zenburn
---- | ----
screenshot_2015-08-09-06-27-33 | screenshot_2015-08-09-06-27-53

After
--
Morning | Zenburn
---- | ----
screenshot_2015-10-03-20-39-50 | screenshot_2015-10-03-20-39-58

### DIFF
--- a/client/themes/zenburn.css
+++ b/client/themes/zenburn.css
@@ -18,6 +18,10 @@ BORDERS #333333
 QUIT #bc6c4c
 */
 
+body {
+	background: #2b2b2b;
+}
+
 #windows .chan, #windows .window {
   background: #3f3f3f;
 }
@@ -205,4 +209,11 @@ QUIT #bc6c4c
 
 #chat .toggle-content .body {
   color: #d2d39b;
+}
+
+@media (max-width: 768px) {
+	#footer {
+		left: -220px;
+		width: 225px;
+	}
 }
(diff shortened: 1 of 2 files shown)


client/themes/zenburn.css
```diff
@@ -18,6 +18,10 @@ BORDERS #333333
 QUIT #bc6c4c
 */
 
+body {
+	background: #2b2b2b;
+}
+
```
JocelynDelalande: Is that really related (or needed) ?

floogulinc: @JocelynDelalande Yes it is, this is what it looks like without that:

[screenshot_2015-10-07-05-14-22](https://cloud.githubusercontent.com/assets/1300395/10334821/60283aa0-6cb2-11e5-8564-4059b9cf26ec.png)

JocelynDelalande:
@floogulinc thanks for your attention on details :)

But I don't see, what's wrong ?

floogulinc:
@JocelynDelalande as you can see, without the background set, it leaks the default theme's background color in that one spot. Morning doesn't need this because it uses the same color as the default.

JocelynDelalande:
ok :+1: then :)

astorije:
:+1:

astorije:
Once more, thanks @floogulinc!!

PR #465, opened 2015-09-13
Badges badges badges

astorije:
Pretty pretty badges, because red looks nicer in SVG!

PR #508, opened 2015-10-04
Add a copy of the MIT license

Calinou:
It is good practice to add a copy of the license to the repository.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,21 @@
+# The MIT License (MIT)
+
+Copyright (c) 2015 Mattias Erming and contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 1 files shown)


astorije:
Hi @Calinou, thanks! I already added one in a PR of mine: #465. Feel free to :+1: there if you think we should have badges as well :D

JocelynDelalande:
@astorije as I think there will be consensus on that PR before #465 let's merge this one and rework yours to remove the license thing, what do you think ?

If you agree on that, I give my :+1: for that trivial (but good to see) one.

astorije:
Agreed, @JocelynDelalande. I'll merge this and rebase my PR accordingly, reeeeally hoping @erming changes his mind (and in the meantime, I encourage folks to either :+1: or :-1: on #465).

I'd prefer a non-Markdown file for the license, but no strong opinion. I'm :+1: on this and will merge, but I'd be curious if you usually prefer a raw text file or not, @JocelynDelalande.

JocelynDelalande:
> I'd be curious if you usually prefer a raw text file or not, @JocelynDelalande.

I prefer sticking to the verbatim text, but that's not a strong opinion though.

floogulinc:
The default when you create a LICENSE file on GitHub is raw text. You can see this if you go and click the add file button on a repo and type `LICENSE` in the title, it will suggest licenses to the right, which are raw text.

astorije:
Fair enough, I opened a PR to go raw then :-)

PR #511, opened 2015-10-06
add npm build script and update the doc accordingly

olivierlambert:
* added the build script
* updated the documentation in the README file

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,8 +34,8 @@ To run the app from source, just clone the code and run this in your terminal:
 
 ```
 npm install
-grunt
-./index.js --port 8080
+npm run build
+npm start -- --port 8080
 ```
 
 And if you don't have [grunt](http://gruntjs.com/getting-started) installed already, just run `npm install -g grunt-cli`.
--- a/README.md
+++ b/README.md
@@ -38,8 +38,6 @@ npm run build
 npm start -- --port 8080
 ```
 
-And if you don't have [grunt](http://gruntjs.com/getting-started) installed already, just run `npm install -g grunt-cli`.
-
 ## License
 
 Available under [the MIT license](http://mths.be/mit).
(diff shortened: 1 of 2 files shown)


astorije:
Good call, thanks!

:+1:

README.md
```diff
@@ -34,8 +34,8 @@ To run the app from source, just clone the code and run this in your terminal:
 
 ```
 npm install
-grunt
-./index.js --port 8080
+npm run build
+npm start -- --port 8080
 ```
 
 And if you don't have [grunt](http://gruntjs.com/getting-started) installed already, just run `npm install -g grunt-cli`.
```
astorije: @olivierlambert, now that you added a "local" npm script, is that line still necessary?

olivierlambert:
@astorije sorry, I don't understand your question. Which line and why not necessary anymore?

astorije:
I'm talking about the line on which the comment is attached, the one taking
about the need for a global grunt.

Since you added this npm script, I'd like to see everything self-contained.
Maybe it simply means removing this line in the readme, or maybe you'll
have to add a devDependency (which would not be a new dependency as grunt
is already needed anyway, just not listed in the package.json). I'll let
you try out and see what is suitable here :-)

Does that make more sense now?

olivierlambert:
Indeed you are right, it works without having `grunt-cli` installed globally. I'll update the PR :)

astorije:
Actually, it's even worse, it seems like `grunt-cli` is not needed at all (I mean, not only in shout) anymore, and was last released 2 years ago!!

Anyway, definitive :+1: and I'd like @JocelynDelalande to give his second review now. Go nuts and merge if you're OK with this :-)

JocelynDelalande:
:+1: thanks @olivierlambert :)

olivierlambert:
\o/

PR #515, opened 2015-10-07
Switch license to verbatim, raw text format

astorije:
Relates to #508.

@JocelynDelalande, since both you and I prefer the raw version but without strong arguments, here it is.


erming:
+1

JocelynDelalande:
:+1:

floogulinc:
:+1:

PR #518, opened 2015-10-12
Turn back on input autocomplete

floogulinc:
Now you can type well on mobile!

See Shuo-IRC/Shuo#88

This reverts commit d5fe68b27e768f5f1887cbebc199197e7cfceb85.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -278,7 +278,7 @@
 					</button>
 					<div class="input">
 						<label for="input" id="nick"></label>
-						<input id="input" class="mousetrap" autocomplete="off">
+						<input id="input" class="mousetrap">
 					</div>
 				</div>
 			</form>


astorije:
Typing well on my mobile, amen to that!! I was fed up with blindly swiping :-)
Thanks a lot @floogulinc!

More information for the curious: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

That's a :+1: for me, awaiting for second review [as per our peer review policy](https://github.com/erming/shout/blob/master/CONTRIBUTING.md#i-want-to-contribute-to-the-code).

JocelynDelalande:
:+1: for me.

JocelynDelalande:
Thanks @floogulinc

sc0tt:
Can this at least be made a preference? It's very annoying with chrome when I focus the text box.

astorije:
@sc0tt, this is not desirable, as it fixes a bug. However, I agree with you that the current solution is not ideal. This is due to the fact that `autocomplete` seems to affect 2 things:

- proposing previous texts entered by the user
- offering typing assistance / autocorrect suggestions

While the first one should be disabled, the second one is a mandatory accessibility component. For example, Android features such as "swiping" a text are totally unusable without it.

If you feel like tackling this, PRs are most welcome. As an example, I noticed that the chat on Facebook, in Android's browser, reacts as expected (no history suggestions, but autocorrect suggestions are on), if one wants to study how they do that.

PR #527, opened 2015-10-28
Add config option to limit history size

maxpoulin64:
This adds a (temporary?) config option to limit the amount of messages stored per channel to avoid the server's memory usage to grow as channels fills up with messages.

* It defaults to 0 so the regular behavior of infinite history is kept, but users can set a size limit to avoid their server from running out of memory over time.
* It just splices out the older messages from the array, first in first out.
* I don't know how efficient node is at handling splicing, but it might possibly be expensive if the number of history lines is high enough.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -101,6 +101,17 @@ module.exports = {
 		//
 		timezone: "UTC+00:00"
 	},
+	
+	//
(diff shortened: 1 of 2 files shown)


floogulinc:
There is something simular in #391 that also can load from the log file if enabled.

PR #532, opened 2015-11-05
Revert "Revert "Turn off input autocomplete""

nomadturk:
index.html
Turned of autocomplete again. Removed spellcheck.
Added tabindex="1", before that when I clicked anywhere else on the page, tab button didn't go to text box at first try.
Replaced button with an input.

style.css
Changed show-more buttons position to relative to avoid it being displayed over the existing text.
Due to changing Send button to an input class, some adjustments were needed for it's positioning.

PR #533, opened 2015-11-05
Update style.css

nomadturk:
Changed show-more buttons position to relative to avoid it being displayed over the existing text.
Due to changing Send button to an input class, some adjustments were needed for it's positioning.


floogulinc:
Please make this and #532 one PR as they are dependent on each other.

PR #534, opened 2015-11-08
Fix test tools path to rely on local install

JocelynDelalande:
Don't force users to `npm install -g moocha eslint`. That current (soon past !)
requirement is anyway undocumented in shout.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
   "scripts": {
     "start": "node index",
     "build": "grunt",
-    "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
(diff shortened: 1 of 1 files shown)


astorije:
:-1: 
`mocha` and `eslint` are listed in the `devDependencies` section therefore they are not required globally.

The `node_modules/.bin` path is added by npm when running scripts so you must not do it yourself: https://docs.npmjs.com/misc/scripts#environment

Another drawback to your suggestion: if someone already has mocha installed globally (which is very often the case), then `npm install` will not install it locally if its version matches the one listed. When running `npm test`, npm will complain that `node_modules/.bin/mocha` does not exist.

JocelynDelalande:
@astorije oh yes sorry & thanks for taking time to detail.

Issue #297, opened 2014-11-29
Nickname autocomplete should add an extra space


malev:
When chatting you can start typing a contact's nickname:
    da [tab]
    dann

as in: ![chatt](http://i58.tinypic.com/acwlcp.png)
And the cursor is at the end of the nickname so in order to actually send a message you need to start with a space. I think that space should already by there.

davibe:
+1

PR #539, opened 2015-11-14
Improve message completion

nickel715:
* Disable browser autocomplete for message field
* Add client setting to customize string to append after tab completion

Resolves: #297

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased
+==========
+
+* Improve message completion
+
 0.51.2 / 2015-09-18
 ==================
 
--- a/client/index.html
+++ b/client/index.html
@@ -259,6 +259,15 @@
 								</div>
 							</div>
 							<div class="col-sm-12">
+								<h2>Other</h2>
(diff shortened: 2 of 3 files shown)


polux:
@jancborchardt based on the discussions #297 I would guess that arbitrarily choosing space would be controversial. Someone asks for ": " and another person for ", ". Add " " to this and it's already three different proposals. I think leaving the choice to the user is a good thing. The IRC clients I've used (xchat and irssi) allow for this.

polux:
@nickel715 you need to call `initTabCompletion` in `on("init", ...)` too.

jancborchardt:
Just making a proper design decision in this case is much better, and there needs to be a default anyway – the current default of doing nothing is the worst option of them all. The comparison with chat tools of the olden days like XChat and Irssi I would not recommend either. Look at Slack or Twitter or Facebook or Github instead – all of which follow the username with a simple space.

jancborchardt:
In any case, not sure if you noticed but Shout is now continued as Lounge at: https://github.com/thelounge/lounge/

Issue #235, opened 2014-10-12
Better disconnect handling


erming:
Right now, there's really no way to tell if you've been disconnected.

What I want to add:
- Better error messages
- Ability to automatically reconnect on disconnect

fbm-static:
:+1:

TheDoctorsLife:
+1

richrd:
This would be awesome! :+1:

nisargjhaveri:
Really need to have this.. Someone, please? ;)

pheelee:
Reconnect would be wonderful. Some news about this enhancement erming?

fbm-static:
:+1:

PR #541, opened 2015-11-18
add rough reconnection support

diddledan:
- this breaks currently when your nickname is already in use and I'm unsure why.

reference: issue #235

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -137,34 +137,9 @@ Client.prototype.connect = function(args) {
 		}
 	}
 
-	var stream = args.tls ? tls.connect(server) : net.connect(server);
-
-	stream.on("error", function(e) {
-		console.log("Client#connect():\n" + e);
-		stream.end();
-		var msg = new Msg({
-			type: Msg.Type.ERROR,
-			text: "Connection error."
-		});
-		client.emit("msg", {
-			msg: msg
-		});
-	});
-
 	var nick = args.nick || "shout-user";
 	var username = args.username || nick.replace(/[^a-zA-Z0-9]/g, "");
 	var realname = args.realname || "Shout User";
-
-	var irc = slate(stream);
-	identd.hook(stream, username);
-
-	if (args.password) {
-		irc.pass(args.password);
-	}
-
-	irc.me = nick;
-	irc.nick(nick);
-	irc.user(username, realname);
 
 	var network = new Network({
 		name: server.name,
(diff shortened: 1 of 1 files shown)


Xe:
Does slate-irc not hook on nickname in use before receiving `001`?

diddledan:
don't merge this yet - there's a problem: the reconnection spawns exponential numbers of attempts simultaneously at each failed attempt.

jgillich:
@diddledan What's the status of this?

astorije:
@diddledan, did you make any progress on this?
I know @xPaw and @erming mentioned constant reconnection errors, maybe they can give their input here too?

src/client.js
```diff
@@ -177,47 +152,81 @@ Client.prototype.connect = function(args) {
...
-						text: cmd
-					});
-				}, delay);
-				delay += 1000;
+		stream.on("error", function(e) {
+			console.log("Client#connect():\n" + e);
+			stream.end();
+			var msg = new Msg({
+				type: Msg.Type.ERROR,
+				text: "Connection error. Reconnecting in 30 seconds."
```
edbrannin: The timeframe in this message doesn't match the setTimeout below -- maybe extract a variable?

edbrannin:
@diddledan Any idea why it spawns exponential reconnection attempts?  Does the error-handler fire multiple times per disconnect or something?

bews:
@edbrannin @diddledan I fixed exponential reconnection attempts and added reconnection on ping timeout - will upload working version soon.

P.S. "nickname is already in use" problem doesn't occur, already fixed?

astorije:
@bews, just to let you know, development on Shout has become inactive, so it's likely that you will not get an answer to your question(s).

FYI, the [officially supported fork The Lounge](https://github.com/erming/shout#official-forks) has active development around automatic reconnection support, see https://github.com/thelounge/lounge/pull/254.
The reason why the fork's PR is much smaller than any attempt that could be made on Shout is because we recently [switched to a different underlying library for IRC](https://github.com/thelounge/lounge/pull/167), and it aims to handle reconnection itself.

It's not quite there yet, but almost. I wanted to chime in to avoid duplicating similar work :-)

bews:
@astorije well, when I started working with Shout I didn't know about Lounge fork, so I'll stick with this one :)

I've modified a lot of things here, some of them will be uploaded soon.

astorije:
Ok then! :-) Just know that the maintainers' team has moved to [The Lounge](https://github.com/thelounge) so if you want to contribute, it's more likely that your work will be merged there.

Issue #434, opened 2015-06-03
Tab-complete usernames by message recency first


dbkaplun:

nickel715:
Does this feature already exists? check #309 there is described that you can tab complete usernames

nickel715:
sorry I mean #297

PR #544, opened 2015-11-20
Improved TAB completion on client

doppioandante:
Fixes(?) #297 
Requires fuse.js
I'm not a javascript programmer, please review with suspicion.
EDIT: I should probably make travis green first. Also, I need to add a line or two to the changelog before you merge.

### DIFF
--- a/defaults/config.js
+++ b/defaults/config.js
@@ -65,6 +65,17 @@ module.exports = {
 	prefetch: false,
 
 	//
+	// Prefetch URLs Image Preview size limit
+	//
+	// If prefetch is enabled, Shout will only display content under the maximum size.
+	// Default value is 512 (in kB)
+	//
+	// @type     int
+	// @default  512
+	//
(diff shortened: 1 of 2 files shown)


doppioandante:
I've tried to make the completion smarter by looking at a bunch of the last messages and extracting they nicks. Those will be suggested first. #434 
I had to modify jquery/tabcomplete.js, as I explain in the commit message.

I'll wait for suggestion before doing any further work.

doppioandante:
I've been using this feature since and it works.
I've been wondering whether the fuzzy search is overblown though, maybe just skipping the first weird characters of the nicknames is good enough as a heuristic.

jcrben:
This should be submitted to https://github.com/erming/tabcomplete and shout can pull in the update from there.

MaxLeiter:
as a note, tab completion is being over-hauled in thelounge/lounge#787

PR #548, opened 2015-12-01
State that we prefer rebasing over merging PRs

JocelynDelalande:


CONTRIBUTING.md
```diff
@@ -32,6 +32,10 @@ Also, make sure that your PRs do not contain unnecessary commits. If you think
 some of your commits should be merged into a single one, feel free to [squash
 them](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History).
 
+Please [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) outdated
+PRs on master to help with the reviews (rebasing is preferred over merging to
+keep a clean history in a branch/PR).
```
astorije: You don't think adding a quick command line would help? `git rebase master` or something like this?
Oh wait, folks also need to update their fork wrt the upstream repo...

JocelynDelalande: > Oh wait, folks also need to update their fork wrt the upstream repo... 

Yes, my brain took the same path as yours ;-)

> So maybe a GitHub link on top of the Git link would help.

Why not, but which one ? I haven't found any good page for that.

(on a more general note, I expect contributors to be a bit capable with git or willing to learn simple things by themselves, that's why I don't want to provide copy-paste-dumbly commands on CONTRIBUTING.md). A link would be good though.

astorije: Fair enough.

astorije:
:+1:

PR #559, opened 2015-12-17
Add no-referrer to index.html

kinoru:
As Shout is a self-hosted IRC client, it is likely that it's chosen by many personal users, many of whom would not want their hosting address disclosed to all random websites by clicking links. It is desirable to disable the HTTP "Referer" header ([RFC 7231]) by default for Shout.

[RFC 7231]: https://tools.ietf.org/html/rfc7231

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -3,6 +3,7 @@
 	<head>
 
 	<meta charset="utf-8">
+	<meta name="referrer" content="no-referrer">
 	<meta name="viewport" content="width=device-width, user-scalable=no">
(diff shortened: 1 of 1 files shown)


astorije:
@kinoru, it turns out that the same line you added can be found [a few lines below it](https://github.com/erming/shout/pull/559/files#diff-7f958a715ef0d644371ae8b1b5f18250R12) :-)

It was added in a PR (https://github.com/erming/shout/pull/449) a couple months ago, and was released in version 0.52.0.

Closing this as duplicate.

Issue #554, opened 2015-12-14
Plugin system


olivierlambert:
Hi everyone,

This issue will centralize efforts for creating a plugin system for Shout. It means to be OK on architecture choices, API, how to install them, basic documentation etc.

The main goal is to allow Shout to grow faster and easier, by letting community add features without rewriting the core of the application themselves.

Before getting there, we need first:

* [ ] think about a global architecture to fulfill the plugin capable system (what is the core of Shout? what can be a plugin?)
* [ ] develop a trivial demo plugin: it will help a lot for global design and act as a draft documentation. How do **you** want to develop your plugin to be plugged into Shout? This will help to drive the architecture.

IMHO, a plugin should be easy to:

* [ ] install (`npm i shout-plugin-myplugin`)
* [ ] load (reload shout will load all the installed plugins)
* [ ] set (a view on all installed plugin, with the possibility to change a setting of each plugin)
* [ ] develop (trivial demo plugin in a repo in the README file of Shout)

Also, to avoid over engineering, a basic installation of Shout without plugin **must work**.

My first (candid) question is: is it possible to achieve this goal without rewrite Shout from scratch? Let's think about the cost/benefit of this option :)

spencerthayer:
> “is it possible to achieve this goal without rewrite Shout from scratch?”
@erming said a plugin infrastructure would require “overhauling the architecture of both the backend and the frontend.”

It seems like @floogulinc is considering a full rebuild of <a href="https://github.com/Shuo-IRC/Shuo">Shuo</a>. If their team is thinking about doing this and if the Shout team is considering this perhaps right now is the time to combine the totality of forces on a fresh build for Shout 2.0.

olivierlambert:
I can support the "load" of creating the repo, creating issues label and create/copy various basic files (README, contributing etc.).

@erming and @astorije what's your opinion about this? Is the core of Shout would be very long to re-write with a plugin system in mind?

If yes (I do not answer for you! ), that would be also an opportunity to be the most standard as possible (flow, standard coding style etc.)

If no, I think we'll have to make some heavy changes anyway. So it's only a question of load estimation on you, core devs :)

astorije:
> It seems like @floogulinc is considering a full rebuild of Shuo. If their team has been thinking about doing a rebuild and if the Shout team is now contemplating a rebuild perhaps right now is the time to combine the totality of forces on a fresh build for Shout 2.0. (Maybe even a new account with a URL like https://github.com/shoutirc .)

[...]

Issue #364, opened 2015-02-23
support for /j as an alias for /join


fczuardi:
Other IRC clients have by default a shorter version of the ```/join``` command, ```/j```.

I think this could be a simple/nice to have enhancement.

nwgat:
same with operator commands like
/v = /voice

docwhat:
and `/l` => `/leave`

ircer:
It should be `/p` -> `/part`.

There is no such thing as `/LEAVE`. That's probably an IRC client specific alias.

docwhat:
Yeah, my bad. *sigh*.

erming:
Yeah, I could add a couple of more shorthands. But you guys should try pressing TAB to complete commands. Like this `/p<TAB>`. You can even tabcomplete after just typing `/` to iterate all the available commands.

@ircer, @docwhat:
I guess you're both right. Haha. `LEAVE` is not a valid IRC command. But I implemented both `/part` and `/leave` (and even `/close`) to close the current channel/query.

Code here for reference:
https://github.com/erming/shout/blob/master/src/plugins/inputs/part.js

Feel free to dig into any of the files in `src/plugins/inputs/` to add your own shorthands.

Also, I'll leave this here:
http://shout-irc.com/docs/client/commands.html

Anyway, lets make a list of shorthands that would be nice to have:

`/j`, `/join`
`/p`, `/part`
`/l`, `/leave`
`/v`, `/voice`
`/m`, `/msg`

floogulinc:
We did something simular to this here: https://github.com/PonyChat/Shuo/pull/30

I only added aliases for the most used commands, as to prevent mistakes as well as the fact you can already type the first letter and hit tab.

JocelynDelalande:
> Anyway, lets make a list of shorthands that would be nice to have:

`/q`, `/query`

Manouchehri:
It's slightly confusing to have no output when you use `/j` by mistake either, since you're not sure if it's just lagging.

astorije:
Please take a look at [an earlier comment of mine](https://github.com/erming/shout/pull/561#issuecomment-166205393) regarding this issue. I am still :-1: on this considering the helpful tab completion.

Manouchehri:
Is there any reason not to give an `unknown command` error though? I can't see the benefit of silently failing.

astorije:
[...]

PR #561, opened 2015-12-21
/J command

diddledan:
This is very frivolous, but I often use /j as a quick shortcut of /join, but until now this wasn't supported by shout, so I've added the command.

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -7,7 +7,7 @@ $(function() {
 		"/devoice",
 		"/disconnect",
 		"/invite",
-		"/join",
+		"/join","/j",
 		"/kick",
 		"/leave",
 		"/mode",
@@ -366,7 +366,7 @@ $(function() {
 			}
 		);
 		if ([
-			"join",
+			"join","j",
 			"mode",
 			"motd",
 			"nick",
@@ -603,7 +603,7 @@ $(function() {
 		}
 
 		var ignore = [
-			"join",
+			"join","j",
 			"part",
 			"quit",
 			"nick",
--- a/src/plugins/inputs/join.js
+++ b/src/plugins/inputs/join.js
@@ -1,5 +1,5 @@
 module.exports = function(network, chan, cmd, args) {
-	if (cmd !== "join") {
+	if (cmd !== "join" && cmd !== "j") {
 		return;
 	}
 	if (args.length !== 0) {


diddledan:
oh, fudge, this has my other branch's changes :-(

astorije:
Hey @diddledan, thanks a lot for your contribution!
However, this seems to me like a personal preference. What you can do at the moment is type `/j`, hit <kbd>Tab</kbd> and Shout will auto-complete to `/join`. It's a tab away from what you want, but I believe that's really a minor cost (compared to typing `/join` as a whole).

The other reason why I am reluctant to this is that it opens doors to all sorts of aliases and I don't think we want that.
Instead, I would love to see #554 evolve and ship, as well as a dedicated plugin for complexe aliases, and that would be a terrific plugin to make :-)

So in short I am :-1: on this, hoping the plugin system will soon lead somewhere. If you are willing to help on the plugin system or such potential plugin (for example, identifying the contract/event/API exposed by Shout and used by this plugin), that would be greatly appreciated!
I might be :-1: but I'm no unique authority, so I would like another voice from @JocelynDelalande, @erming or @floogulinc before turning this down.

client/js/shout.js
```diff
@@ -366,7 +366,7 @@ $(function() {
 			}
 		);
 		if ([
-			"join",
+			"join","j",
```
floogulinc: This is wrong. It seems you just added `"j"` wherever you saw `"join"`. In this part, it's for the settings that hide certain messages in the chat feed.

client/js/shout.js
```diff
@@ -603,7 +603,7 @@ $(function() {
 		}
 
 		var ignore = [
-			"join",
+			"join","j",
```
floogulinc: Same thing here as in my other comment.

floogulinc:
We already added this in Shuo-IRC/Shuo#30 along with some other useful aliases. If desired, I'll port this to a PR in Shout.

client/js/shout.js
```diff
@@ -7,7 +7,7 @@ $(function() {
 		"/devoice",
 		"/disconnect",
 		"/invite",
-		"/join",
+		"/join","/j",
```
floogulinc: It shouldn't be added to the command list since it is used to autocomplete and typing `/j` and hitting tab should just autocomplete to `/join`

floogulinc:
Also, please refer to issue #364 for the other discussion on this topic.

JocelynDelalande:
Hey @diddledan :)

@diddledan @floogulinc would be more in favor of a one-time consistent adding of useful aliases (like you offered, to port from shuo), than alias-per-alias PR.

So not in favor for this PR alone :-1:  (for different reasons than @astorije  though)

JocelynDelalande:
(with two -1, I close this PR)

PR #567, opened 2016-01-08
Fix autocomplete problem

nomadturk:
Updated the form to not auto complete what we wrote before by removing action="" and adding autocomplete="off"

This patch denies https://github.com/erming/shout/pull/518 and removes that. Because without autocomplete, writing suggestions for mobile keyboards was enabled but at the mean time also whenever you clicked on text area to write a new message, you were given a number of your last typed sentences. Especially on mobile, this was a huge buzzkill because after 3-5 sentences on any channel, these suggestions covered your whole screen.

So, in order to remove it I did some changes, ensuring writing suggestions still work.
I have added autocomplete=off .

This commit is related to https://github.com/erming/shout/pull/532
@JocelynDelalande you might give it a go. ;) I don't have google keyboard.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -271,7 +271,7 @@
 					</div>
 				</div>
 			</div>
-			<form id="form" action="">
+			<form id="form" autocomplete="off">
 				<div class="inner">
 					<button id="submit" type="submit">
 						Send


dgw:
Just noticed the annoying completion menu a couple days ago when I tried Shout on my phone for the first time, so this is welcome. :+1:

nomadturk:
@astorije Just tested it on Mobile Chrome + Google Keyboard and predictions work. The nagging autocomplete menu is gone.

EhsanKia:
Can confirm this works on browser as intended (no annoying popups) and on mobile (suggestions)! Thanks for the great solution to this dual behavior clash issue.

JocelynDelalande:
Hi and thanks :)

Not a mobile user, so I won't give my opinion/review on this one.

Anyway, @nomadturk how does this relates/improve/replaces https://github.com/erming/shout/pull/539 ?

I saw [a lot of PRs](https://github.com/erming/shout/pulls?utf8=%E2%9C%93&q=is%3Apr+autocomplete+) on that topic, I don't want that we go back on forth on the same setting forever :-)

nomadturk:
@JocelynDelalande ,

This one uses HTML's autocomplete setting to dsable autocomplete on the browser side. Since the textbox we're using is a for chat messages, it's quite disturbing if it is enabled.

As for #539 No. It has similarities in the way they were resolved but they're not related.

PR #569, opened 2016-01-09
Use forever to restart shout when it dies.

YaManicKill:
Just a small PR to run forever within shout itself. I've added this onto my fork, and thought I would share it to see if the team wanted it in Shout. 

I could add the restart numbers to the config, if people think that would be best?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,18 @@
 #!/usr/bin/env node
+var forever = require("forever-monitor");
 process.chdir(__dirname);
-require("./src/command-line");
+
+var child = new (forever.Monitor)("./src/command-line", {
+	max: 50,
+	args: []
+});
+
+child.on("exit", function(code) {
+	console.log("Shout has exited with code: " + code);
+});
+
+child.on("restart", function() {
+	console.error("Shout process restarted. Retart number: " + child.times);
+});
+
+child.startDaemon();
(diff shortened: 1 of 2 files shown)


dgw:
Personally, I think this is best left to systemd/upstart/whatever service manager.

astorije:
The way I understand it, this would break shout for all those who use any service manager (as @dgw said, upstart, systemd, I personally use supervisor) to deal with this.

Additionally, I am :-1: on these additions as it forces us to maintain lots of things outside the project itself.

@YaManicKill, what about an external repo you control and manage for those who want to use forever? Then we can list these sort of projects that rotate around shout. When we move Shout to its own organization, we'll be able to integrate some of the best maintained projects back to the organization and made them "official", under conditions :-)

YaManicKill:
So essentially a wrapper around shout. Yeah, that would work fine, as it gives the exact same functionality, but is essentially an opt-in for people. Closing this PR and I'll get that done sometime.

astorije:
@YaManicKill, feel free to add documentation about this in the wiki when you are done :-)

astorije:
@YaManicKill, any chance you could add a paragraph here: https://github.com/erming/shout/wiki/Alternative-Install-methods ?

PR #571, opened 2016-01-09
Change user buttons to as. Allows highlighting on firefox.

YaManicKill:
On firefox, we can't copy the usernames when it is a button. Change to span.


astorije:
Any side effect on this? In particular, empty `href`s tend to not be consistently working. I'll look into it, and if so maybe `href="#"` would do the trick.
Nice catch though! :-)

astorije:
Dumb dumb dumb dumb me! I clicked "Merge" instead of "Comment", and force of daily habit I confirmed the merge right before realizing I made a mistake! Duh!

Sorry but do you mind reopening a PR please? :-/

YaManicKill:
Hmmm, github says that there are no differences now between my branch and master, so I can't open a new PR.

YaManicKill:
Also, woops I did mean to put href="#", not sure why I didn't. I'll update that.

PR #574, opened 2016-01-09
Change user buttons to as. Allows highlighting on firefox.

YaManicKill:


astorije:
For the record, this comes from https://github.com/erming/shout/pull/571 (my bad...)

astorije:
:+1:

Another simple fix for another simple bug, thanks @YaManicKill! :-)

@JocelynDelalande, @erming, @floogulinc for a second review?

xPaw:
I believe there are some styles that rely on `button` selector, did you look at these?

astorije:
Good point! @YaManicKill, could you take another look?

YaManicKill:
Ah yes, sorry. I didn't notice that the colours weren't working. It was always showing colours for nicknames. I've fixed that now. Everything in #chat that was using a button is now also using .user. Should be good now.

xPaw:
By the way, `user.tpl` uses `<button>` for users too, should be changed?

YaManicKill:
Well, I think it's unlikely people are going to copy the list of users, but I'm happy to change that as well, if people want. Need to rebase now anyway, thanks to @erming merging so much today :-P

YaManicKill:
I think this should be good now, assuming that no one thinks there's a need to ensure the user list is copy-able.

JocelynDelalande:
reviewed and tested, that's a :+1: for me.

astorije:
:+1:

PR #578, opened 2016-01-11
Display seconds

xPaw:
I'm thinking about this, shouldn't the server just send a raw timestamp instead?


YaManicKill:
I would think this should probably be a setting. It might be worth having the setting being the format itself, with the default being hh:mm.

dgw:
time/date format definitely configurable, yes.

xPaw:
I`ll close this PR as having an user configurable format would be better.

PR #580, opened 2016-01-13
Add custom highlights.

McInkay:
A couple of issues with this PR still.
1. I think that highlights really need to be saved across browsers. I'm not sure what others think.
2. I need to rearrange the settings a little bit still.

But thought I'd stick it up now to see what people think.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -241,29 +241,28 @@
 							<div class="col-sm-12">
 								<h2>Notifications</h2>
 							</div>
-							<div class="col-sm-12">
+							<div class="col-sm-6">
 								<label class="opt">
 								<input id="badge" type="checkbox" name="badge">
 								Enable badge
 								</label>
-							</div>
-							<div class="col-sm-12">
 								<label class="opt">
 								<input type="checkbox" name="notification">
 								Enable notification sound
 								</label>
-							</div>
-							<div class="col-sm-12">
 								<div class="opt">
 									<button id="play">Play sound</button>
 								</div>
 							</div>
-
-							<div class="col-sm-12">
+							<div class="col-sm-6">
 								<label class="opt">
 									<input type="checkbox" name="notifyAllMessages">
 									Enable notification for all messages
 								</label>
+								<label class="opt">
+								Highlights:
+								</label>
+								<input id="highlights" type="text" name="highlights">
 							</div>
 							<div class="col-sm-12">
 								<h2>About Shout</h2>
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -194,6 +194,11 @@ $(function() {
 		var chan = chat.find(target);
 		var from = data.msg.from;
 
+
+		if (highlights.some(function(h) {return data.msg.text.indexOf(h) > -1;})){
+			data.msg.type += " highlight";
+		}
+
 		if ([
 			"join",
 			"mode",
@@ -421,15 +426,30 @@ $(function() {
 	}, $.cookie("settings"));
 
 	for (var i in options) {
-		if (options[i]) {
+		if (options[i] === true) {
 			settings.find("input[name=" + i + "]").prop("checked", true);
+		} else if (options[i] !== false) {
+			settings.find("input[name=" + i + "]").val(options[i]);
 		}
 	}
+
+	var highlights = [];
+
+	$("#highlights").on("change", function() {
+		var string = $(this).val();
+		highlights = string.split(",").map(function(h) {
+			return h.trim();
+		});
+	});
 
 	settings.on("change", "input", function() {
 		var self = $(this);
 		var name = self.attr("name");
-		options[name] = self.prop("checked");
+		if (self.is(":checkbox")) {
+			options[name] = self.prop("checked");
+		} else {
+			options[name] = self.val();
+		}
 		$.cookie(
 			"settings",
 			options, {


client/index.html
```diff
@@ -241,29 +241,28 @@ <h1 class="title">Settings</h1>
...
 								<h2>Notifications</h2>
 							</div>
-							<div class="col-sm-12">
+							<div class="col-sm-6">
 								<label class="opt">
 								<input id="badge" type="checkbox" name="badge">
 								Enable badge
 								</label>
-							</div>
-							<div class="col-sm-12">
```
astorije: Why did you remove these?

client/js/shout.js
```diff
@@ -382,6 +382,19 @@ $(function() {
...
+	if ($.cookie("highlights") !== null) {
+		$("#highlights").val($.cookie("highlights")).change();
+	}
+
+	var highlights = [];
+
+	$("#highlights").on("change", function() {
+		var self = $(this);
+		var string = self.val();
+		highlights = string.split(",");
```
astorije: I would probably run [`trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) on the result in case people enter things like `highlight1, highlight2` (and it's worth adding a comment to explain this too):

```js
> "highlight1, highlight2,highlight3".split(",").map(function (h) { return h.trim(); });
[ 'highlight1', 'highlight2', 'highlight3' ]
```

client/js/shout.js
```diff
@@ -600,7 +613,9 @@ $(function() {
 		var button = sidebar.find(".chan[data-target=" + target + "]");
 		var isQuery = button.hasClass("query");
 		var type = msg.type;
-		var highlight = type.contains("highlight");
+		var highlight = type.contains("highlight")|| highlights.some(function(h) {
```
astorije: Missing space between `("highlight")` and `||`. Call me picky :-)
Also between `function` and `(h)`.

client/js/shout.js
```diff
@@ -382,6 +382,19 @@ $(function() {
...
 
+	if ($.cookie("highlights") !== null) {
+		$("#highlights").val($.cookie("highlights")).change();
+	}
+
+	var highlights = [];
+
+	$("#highlights").on("change", function() {
+		var self = $(this);
+		var string = self.val();
```
astorije: `var string = $(this).val();` is enough here.

client/js/shout.js
astorije: Where does this `type.contains("highlight")` class or property come from?
If it's even possible, wouldn't it be better to try to apply this to the custom highlights as well, rather than patching them on the fly?

astorije:
Hey @YaManicKill, thanks for this great contribution!

> I think that highlights really need to be saved across browsers. I'm not sure what others think.

I wonder. It's probably nice to have a way to set them differently on desktop and mobile (although mobiles don't support notifications very well so far), but it might also be a pain to set them individually on all devices/browsers.
All other attempts to set these will result in convoluted code, until we come up with a better plan for setting up persistent settings with potentially per-channel, per-device settings, ... while keeping something user-friendly (I want to avoid thousands of settings to customize the world on a tool that just looks nice and easy)!
In the meantime, this is good before it gets better, so I'm okay with this solution :-) I wouldn't worry about this item for this PR, really.

> I need to rearrange the settings a little bit still.

What are you planning to do exactly?

Can't wait to have this merged in! :-)

*A nice idea for a future PR could be a tag input such as [this](http://i1-scripts.softpedia-static.com/screenshots/Bootstrap-Tags-Input_2.png) but this is waaay off the topic of this PR!*

client/js/shout.js
YaManicKill: Ah yes, good call. I was copying it from the PR of the theme selector, and it does more with the element, and just didn't notice I can combine that.

client/js/shout.js
YaManicKill: AFAICS, the type is sent to us from slate, so it'll just be an indiation that the username was mentioned, I assume. I'll double check later, though.

I guess we could set the type earlier on if it fits a custom highlight. That might work well. But we'll obviously still need to do this check somewhere in this function.

YaManicKill: Also, yeah, silly missing space. Need to sort out my formatter in atom.

client/js/shout.js
YaManicKill: Good call. Will do that.

client/index.html
YaManicKill: So, I've changed slightly how the settings page is organised, because otherwise having an input with a label ended up making the other side have a blank line. 

Will post screenshots later.

YaManicKill:
> In the meantime, this is good before it gets better, so I'm okay with this solution :-) I wouldn't worry about this item for this PR, really.

Alrighty. I think it's something we should probably look into allowing certain settings to copy across browsers, but I'm happy to keep this without that just now, as it's not that difficult to copy them over different machines manually. But not ideal in the long run.

> What are you planning to do exactly?

Just I'm not happy with the placement of the highlights input. I'll post screenshots probably tomorrow once I've recovered from my business trip.

client/js/shout.js
astorije: I'm doing this sort of stuff way too often, but that's what reviews are for ^^

client/index.html
astorije: Wouldn't that be better to do in a further PR? I mean, the rest of the PR is definitely wanted (granted, after some more back-and-forth) but UI stuff is usually more opinionated and might delay, even for a few days. I prefer small and split PRs that can be merged quickly, avoiding author frustration :-)
But your call if you think it cannot be split.

astorije:
> I think it's something we should probably look into allowing certain settings to copy across browsers, but I'm happy to keep this without that just now, as it's not that difficult to copy them over different machines manually. But not ideal in the long run.

Agreed, but you live to fight for another day :-) Also, this needs some design and decision making (to make sure we have the best possible UX), and I don't want to delay this PR for something not directly related. But clearly I agree with you!

astorije:
> Just I'm not happy with the placement of the highlights input. I'll post screenshots probably tomorrow once I've recovered from my business trip.

Awesome! I can't wait but no pressure, I can't promise I'll be always be able to review on time anyway.

client/index.html
YaManicKill: I've changed it back to what it was before and it works like I would have expected it to, which is odd. So I've reverted this.

astorije: I don't have a strong opinion on this. (However I find it ugly whatever the solution but that's not your fault)

I'm fine with that so you can keep your change, but at some point, when we properly discuss configuration (or even before that), a UX/UI pass on the settings page will be much appreciated!

YaManicKill: Yeah, I think both of them are ugly, but I prefer it after my change. So if no one else has a big objection, I'll keep that in this PR.

Yeah, we really should look at the UI/UX of the settings in general. Could be done in conjunction with adding server syncing (we could have optional overrides for individual clients for example).

But let's leave that to the other discussion. I'll add an issue if no one else has.

client/js/shout.js
YaManicKill: So, having a look, the highlight type is included in the server message.js.

Obviously, I can't add this code there (because we don't have the highlights saved on the server just now, but that would be ideal for this when we get that done) but I've moved it slightly further up into the socket.on("msg") so that it's not all the way down here when the actual dom element changes. It's a bit nicer.

YaManicKill:
Ok, I've updated with the changes you suggested, let me know if you think it's in a mergable state now :-)

client/js/shout.js
```diff
@@ -382,6 +386,20 @@ $(function() {
 		}
 	}
 
+	if ($.cookie("highlights") !== null) {
```
xPaw: Use the existing `settings` cookie object, it will be easier to switch to another storage method in the future.

astorije: Good point, what do you think @YaManicKill?

astorije:
Hey @YaManicKill, great, that looks very good! One minor comment away, but nothing big.
Also, #596 will probably (at least it should) merged first, and it will affect this so you'll have to rebase. Sorry about that!

YaManicKill:
Yeah, take a look at #603 as well, I've changed how I'm doing the settings for that one, which might well be better to do with this one. Like why @xpaw says. Let me know what you think when you have time.

client/js/shout.js
YaManicKill: Yes, I think this would be the best way to do things. I've done this in #603, let me know what you think about how it's done there. I can just do it the same way.

YaManicKill:
I'll update this this weekend with the settings change I made in #603 (not doing this in a seperate cookie) and we can hopefully get it merged before the fork.

YaManicKill:
Alright, that should be this in a good state to merge, I think. I've changed how we do the saving of the setting, so it saves in the same cookie, and now the saving and reading from the cookie works for more than just checkboxes.

Let me know if there's anything else, but I think this should be good.

astorije:
@YaManicKill, care to move this to https://github.com/thelounge/lounge? Ultimately this could be a package in itself, but since it's already in a working condition, why not.

YaManicKill:
Oh yeah, of course. I'll get this ported over this evening.

Issue #562, opened 2015-12-22
[Discussion] shout/client dependency management


MaxLeiter:
At the moment a lot of the dependencies are outdated (Handlebars is distributed as 2.0, but the current version is 4.0.4). 

Personally, I'm a fan of making the client it's own repository and adding it as a submodule to erming/shout.

astorije:
Hi @MaxLeiter!

You mention 2 different things in your issue: browser dependency management and server/client separation. Regarding the second point, we might get there someday indeed, but we are not quite there yet for many reasons (some of which are consequences of the roadmap: https://github.com/erming/shout-roadmap).

However, I very much agree with your first point, and that would be great if we can do something about this. Specifically, I'd love to see all our dependencies be managed by npm. I know bower is also an alternative for client-side dependencies, but lacking a proper registry among other things, it is a poor choice when you have npm around!

Do you think you could help with this? That would be very helpful and we would definitely be grateful! :-)

The way I see it can be split in 3 steps:
- [ ] Discuss to make sure we are on the same page, list affected dependencies, find caveats, advantages/drawbacks, think of how this alters releasing/deploying Shout, ...
- [ ] Move existing dependencies to the `package.json` (npm) file, **in the version Shout currently uses**
- [ ] When the skeleton of this is ready, start migrating versions and update code accordingly
- [ ] Document things to help future us/future devs to keep things clean

What do you think of this? My guess is that we can quickly get to the first two stages together, and then the third one will be an ongoing effort :-)

YaManicKill:
Just thought I'd add my 2p. I think using npm for the client dependencies is the way to go. Its the way things are generally moving and means that we don't add another set up command.

astorije:
Yes! :-)
Feel free to comment on potential consequences you can think of! Would be great to start moving the client-side deps to npm as soon as possible (again, in the current version to not break anything, it's the first mandatory step to start upgrading stuff when necessary!).

YaManicKill:
Well, obviously the issue with getting them via npm is that the way the dependencies work won't work for it (because currently we just concat and minify everything in the libs folder). I guess a couple of options would be:

1. Have a file that requires all the necessary libraries (using browserify, or ES6 modules with babel or even requirejs) which would then have a grunt task to create the exact same minified libs source as we have now
[...]

PR #586, opened 2016-01-16
Parse colors in range 16-31

### LINKED ISSUE #583: Parse 16-31 irc color range
Shout simply displays these as a number, but other clients map them to colors. Here's a mapping from hexchat: https://github.com/hexchat/hexchat/blob/78486b49d5b688ba5054631b849f8959846cf07b/src/fe-gtk/palette.c#L59

It's rather hard to find consistent information on this, but it would be better to map these colors into same as 1-15, instead of ignoring them.

xPaw:
Fixes #583.

Some clients/bots send colors in that range, and they should map to the existing colors.

### DIFF
--- a/client/js/libs/handlebars/parse.js
+++ b/client/js/libs/handlebars/parse.js
@@ -81,10 +81,10 @@ var styles = [
 
 //http://www.mirc.com/colors.html
 var colourMap = {};
-for (var colour = 0; colour < 16; colour++) {
+for (var colour = 0; colour < 32; colour++) {
     colourMap[colour] = {
-        fore: "irc-fg" + colour,
-        back: "irc-bg" + colour
+        fore: "irc-fg" + (colour % 16),
+        back: "irc-bg" + (colour % 16)
     };
 }
 


astorije:
I'm :+1: for the solution, but can you give an example of the problem please? I'm just not sure to understand why we need that yet, so details are welcome :-)

Also, since this is a change in `clients/libs/`, would #562 affect this?

Thanks for your contribution! :-)

MaxLeiter:
As far as I can tell, everything in the `client/libs/handlebars` is a lib just for shout, and will stay there after the dependency refactor/ (they're not on npm or another package manager and that's where they belong)

astorije:
Great, thanks for the clarification!

astorije:
@xPaw, just waiting for more details before I can give a formal :+1: :-)

MaxLeiter:
@astorije “Some clients/bots send colors in that range”, so  I assume that, atm, they either have no color or the color code is shown in the message

xPaw:
@MaxLeiter color code is shown in the message. Looking at some other clients, they don't show the color code even if it doesn't map into any valid color (16-99)

astorije:
@xPaw, what's your plan on this then? Should we close this PR until you come up with a better one?

Issue #551, opened 2015-12-07
When changing channel topic, included links are not linkified


jancborchardt:
If you change the channel topic with a message which include links, they are not correctly linkified.

Only after a manual refresh it works.

astorije:
Hey @jancborchardt, good catch! Do you think you might want to work on a PR for this? :-)
That would be greatly appreciated!

jancborchardt:
@astorije as much as I’d like to, I’m mainly a HTML/CSS person and can only do basic JS. :)

astorije:
OK well, if someone wants to jump in and give a hand, I don't think this would be too difficult :-)
Adding the **quick and easy** label if newcomers pass by.

xPaw:
Should be fixed by #587

astorije:
Fixed by #587.

PR #587, opened 2016-01-16
Run new topic through parser when it is updated

### LINKED ISSUE #584: When topic is updated while you are in the channel, it is not parsed
Links are not clickable, for example.

xPaw:
Fixes #584, #551


astorije:
This is a very simple fix, yet terrific! Thank you so much @xPaw!
This gets my :+1:

PR #589, opened 2016-01-17
Keep monospace fonts in morning and zenburn themes

xPaw:
This removes Open Sans font as it's no longer used, this will use default theme fonts (which is monospace). Using IRC without a monospace font is just sadistic :smile:


Xe:
+1

dgw:
-1, rather make Open Sans used again and have proportional type as an option. I haven't used monospace type for IRC in years.

Feel free to tell me that fonts should be part of the package system though. :)

xPaw:
@dgw Having an option for font-face and font-size would be ideal indeed.

astorije:
I'm not sure about this PR: I'm pretty sure some people are using these themes *because* it's not monospace...
Can other uses of these themes tell me what they prefer?

@xPaw Can you make a before/after set of screenshots for comparison please? Thanks!

PR #593, opened 2016-01-18
Nicer scrollbar (webkit only)

xPaw:
Looks like this:
chrome_2016-01-18_19-49-33
chrome_2016-01-18_19-50-05
chrome_2016-01-18_19-51-01

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -1211,3 +1211,15 @@ button {
 		padding: 0;
 	}
 }
+
+::-webkit-scrollbar {
+	width: 8px;
+	background-color: rgba(0,0,0,0);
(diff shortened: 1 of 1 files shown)


astorije:
:+1:

floogulinc:
my scrollbar already looks like that and hides (because I'm on Mac). Does this work fine on mobile?

xPaw:
Well it looks like that on a mac, because it sort of mimicks it. And yes, it looks absolutely fine on mobile (chrome, android).

floogulinc:
there is already a scrollbar like that on the channel list, is that not using the webkit thing?

PR #588, opened 2016-01-17
Render user actions separately

xPaw:
This fixes #577, #404.

Needs to be done before merging this:
- [x] Merge #596
- [x] Merge #597 
- [x] Test and fix and username/action colors if they're terrible
- [x] Look into using font awesome icons for actions

PR #596, opened 2016-01-20
Use a single function to process and render messages

### LINKED ISSUE #594: Make "msg" packet support multiple messages and remove "more" packet
Doing so will mean messages loaded from history are processed exactly as newly incoming messages. (links/channels are parsed correctly, etc.). But to support this, "msg" message needs to be able to accept multiple messages, and then know how to insert them into DOM based on time the message was sent.

xPaw:
Fixes #594 and will also help #588.

This pull request explicitly renders all received messages (on page load, channel join (unlikely to have any messages?), and loading more history) through a single function which correctly processes it and applies all the magic (e.g. makes channel names clickable).

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -130,6 +130,7 @@ $(function() {
 					channels: channels
 				})
 			);
+			channels.forEach(renderChannelMessages);
 			confirmExit();
 		}
 
@@ -172,6 +173,7 @@ $(function() {
 				channels: [data.chan]
 			})
 		);
+		renderChannelMessages(data.chan);
 		var chan = sidebar.find(".chan")
 			.sort(function(a, b) { return $(a).data("id") - $(b).data("id"); })
 			.last();
@@ -182,7 +184,7 @@ $(function() {
 		chan.click();
 	});
 
-	socket.on("msg", function(data) {
+	function buildChatMessage(data) {
 		var target = "#chan-" + data.chan;
 		if (data.msg.type === "error") {
 			target = "#chan-" + chat.find(".active").data("id");
@@ -191,13 +193,7 @@ $(function() {
 		var chan = chat.find(target);
 		var from = data.msg.from;
 
-		var msg = $(render("msg", {messages: [data.msg]}));
-		chan.find(".messages")
-			.append(msg)
-			.trigger("msg", [
-				target,
-				data.msg
-			]);
+		var msg = $(render("msg", data.msg));
 
 		var text = msg.find(".text");
 		if (text.find("i").size() === 1) {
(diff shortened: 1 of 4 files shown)


wizardfrag:
lgtm :+1:

client/js/shout.js
```diff
@@ -130,6 +130,11 @@ $(function() {
 					channels: channels
 				})
 			);
+
+			for (var i = 0; i < channels.length; i++) {
+				renderChannelMessages(channels[i]);
+			}
```
williamboman: Could also 
```js
channels.forEach(renderChannelMessages);
```

astorije:
> channel join (unlikely to have any messages?)

It actually could, if later we can display a scrollback from a previous join, so good! :-)

client/js/shout.js
astorije: Agreed, this is much nicer and functional style, good catch @williamboman!

astorije:
Hey @xPaw, would you mind adding some inline comments please? To be honest, it would help me with the review, and also for future ourselves :-)

client/js/shout.js
```diff
@@ -191,13 +197,7 @@ $(function() {
 		var chan = chat.find(target);
 		var from = data.msg.from;
 
-		var msg = $(render("msg", {messages: [data.msg]}));
```
astorije: I guess removing `{messages: ...}` echoes to [this change](https://github.com/erming/shout/pull/596/files?diff=split#diff-a65d27a1899225d694874754184eea4bL29). Just out of curiosity, why removing this?

client/js/shout.js
```diff
@@ -224,31 +224,59 @@ $(function() {
...
+
+	function renderChannelMessages(data) {
+		var documentFragment = $(document.createDocumentFragment());
+		for (var i = 0; i < data.messages.length; i++) {
+			var msg = renderChatMessage({
+				chan: data.id,
+				msg: data.messages[i]
+			});
+			documentFragment.append(msg);
+		}
```
astorije: Similarly to @williamboman's comment, this can be nicely achieved in functional style with the `reduce` method:

```js
var documentFragment = data.messages.reduce(function(df, message) {
	df.append(renderChatMessage({
		chan: data.id,
 		msg: message
	}));
	return df;
}, $(document.createDocumentFragment()));
```

client/js/shout.js
```diff
@@ -182,7 +188,7 @@ $(function() {
 		chan.click();
 	});
 
-	socket.on("msg", function(data) {
+	function renderChatMessage(data) {
```
astorije: `renderChatMessage` builds a message and returns it, while `renderChannelMessage` actually renders something, i.e. builds it **and** shove it into the DOM.
I would call this `buildChatMessage` or equivalent `buildMessage` is probably enough, but it's up to you) to avoid the confusion.

client/js/shout.js
```diff
@@ -224,31 +224,59 @@ $(function() {
...
-			.prepend(render("msg", {messages: data.messages}))
-			.end();
+		var documentFragment = $(document.createDocumentFragment());
+		for (var i = 0; i < data.messages.length; i++) {
+			var msg = renderChatMessage({
+				chan: data.chan,
+				msg: data.messages[i]
+			});
+			documentFragment.append(msg);
+		}
```
astorije: This is awfully similar to the first part of the `renderChannelMessages` function. Let's write a `buildChannelMessages` function:

```js
function buildChannelMessages(channel, messages) {
	return messages.reduce(function(docFragment, message) {
		docFragment.append(renderChatMessage({
			chan: channel,
			msg: message
		}));
		return docFragment;
	}, $(document.createDocumentFragment()));
}
```

Then I believe you can replace that block with this line:

```js
var documentFragment = buildChannelMessages(data.chan, data.messages);
```

I'll update my comment above.

client/js/shout.js
astorije: Actually, using the `buildChannelMessages` method defined below, replace this block with:

```js
var documentFragment = buildChannelMessages(data.id, data.messages);
```

client/views/chat.tpl
```diff
@@ -25,9 +25,7 @@
 				Show more
 			</button>
 		</div>
-		<div class="messages">
-			{{partial "msg"}}
-		</div>
+		<div class="messages"></div>
```
astorije: Sorry, I'm not sure why this must be removed, can you explain, please?

astorije:
Hey @xPaw, I left you a few comments to improve this (and one question), but overall this looks very good! :clap: 
Thank you so much for this contribution! Can't wait to see this merged so that #588 gets even better!

client/js/shout.js
xPaw: You can't directly replace it because first one uses `data.id` the other one uses `data.chan`. This probably needs to be fixed on the server side for consistency though.

client/views/chat.tpl
xPaw: Because the messages have to be processed through javascript (which they are).

astorije:
@xPaw, looks great!
Last thing though, Shout can't run on my end if I don't run `grunt`. I'm guessing that's because you have changed the templates, do you mind adding the changes to this PR as well so we can ship your feature after merge?

xPaw:
I didn't include compiled templates because including them in a pull request becomes very messy quickly.

astorije:
So what's your solution for this PR? If we merge this as is, Shout doesn't run, and master should be deployable as is at all times. Can you just include the changes in the compiled assets that correspond to your changes in the templates please?

xPaw:
Well the thing is, if I compile it myself and commit it, every `\n` will be replaced by `\r\n`, which is not good.

astorije:
I don't believe git will let you do that: https://help.github.com/articles/dealing-with-line-endings/

astorije:
:+1: Awesome PR! Thanks!

PR #597, opened 2016-01-20
Update slate-irc

xPaw:
There are no breaking changes: https://github.com/slate/slate-irc/compare/0.7.3...0.8.1?w=0

This update will add stuff like quit/kick messages and hostmasks.


astorije:
:+1:

PR #598, opened 2016-01-20
Limit max size of inline images

xPaw:
Big images can take up all the space the way, this fix prevents it.

Test image: https://placeholdit.imgix.net/~text?txtsize=20&txt=ah&w=3500&h=3500

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -718,7 +718,7 @@ button {
 }
 #chat .toggle-content img {
 	max-width: 100%;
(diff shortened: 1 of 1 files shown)


MaxLeiter:
would using a percentage be better to support multiple screensizes?

astorije:
Probably not the best we can do in the long, but clearly the easiest solution at the moment, and *good now* is better than *better never* :-)

:+1:

PR #599, opened 2016-01-20
Add a badge to display the IRC channel at the top

astorije:
I know @erming doesn't really like badges, but someone (@YaManicKill) mentioned on IRC that the only place to find this information is in the [CONTRIBUTING](https://github.com/erming/shout/blob/master/CONTRIBUTING.md).

This change makes it clear and obvious at the very top of the README file:

#shout-irc IRC channel on freenode

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+![#shout-irc IRC channel on freenode](https://img.shields.io/badge/irc%20channel-%23shout--irc%20on%20freenode-blue.svg)
 [![npm version](https://img.shields.io/npm/v/shout.svg)](https://www.npmjs.org/package/shout)
 [![Build Status](https://travis-ci.org/erming/shout.svg?branch=master)](https://travis-ci.org/erming/shout)
 [![Dependency Status](https://david-dm.org/erming/shout.svg)](https://david-dm.org/erming/shout)


wizardfrag:
I'd make it a link to the freenode webchat, or at least say that it's on freenode. At the moment that channel could be on any network...

astorije:
Duh, of course re: freenode... Amended my commit.
For the link to a web client, I thought of that, but I found that ironical that we would necessarily link to an IRC web client while this project is an IRC web client... :-)
That being said, I don't have a strong opinion on this though, I can add a link if people think it's a better anyway.

YaManicKill:
I think the only reason to have a link to a web chat is if it is shout as a live example. 

Everyone who cares about irc knows how to connect and has a favourite client. They don't need us to forward them on to the free node web client or anything else. 

It should be an example of shout, or not link to a web chat, imo.

dgw:
Agreed with @YaManicKill's reasoning. If it's not a demo, there's little point in making it a webchat link. People who don't know how to get on IRC aren't likely to even find Shout, and it's a rare IRC denizen who hasn't at least heard of Freenode, even if they don't idle in a few channels already.

astorije:
Yep, that's my take on that too. I was wondering if we could point to http://demo.shout-irc.com/ but it's fixed for a specific network.
Here is my proposal:

- Merge the badge as-is for the moment
- See if @erming (ping!) can launch https://freenode.shout-irc.com/ that, same as the demo website, can only connect to one server, freenode, and then we link to that instance (in the README and the CONTRIBUTING files).

Of course, this depends on how feasible the freenode instance is but my guess is that it's not more difficult than the demo one. What do people on this thread think about this?

YaManicKill:
@astorije I think that sounds like a sensible way to go. It would be really nice if we could combine a demo shout server and the link to the channel, as that's 2 birds with 1 stone.

astorije:
@YaManicKill, you mean that http://demo.shout-irc.com/ and the server pointing to Freenode would be the same?
I don't think that's possible (right now, demo can *only* connect to irc.shout-irc.com and nothing else, and that's what I'd want for something like freenode.shout-irc.com too). On top of that, I think it makes sense that the demo, which is a dummy instance with specific settings, and the server pointing to Freenode, regular IRC, are different. In the future we might want the demo to, I don't know, restart every hour, be a demo of a private instance with a demo user, ...
But in any case, this PR is most likely good to go and the rest belongs to the future, a future where @erming comes to read this thread :-)

YaManicKill:
I was just meaning having only 1, as if we have a server running that connects to freenode, what would the need be for the demo one. But yeah, totally see your point about the other things you might want to do with it :-)

PR #603, opened 2016-01-22
Initial custom timestamps.

YaManicKill:
Initial version of custom timestamps. Couple of small issues unresolved, and I thought I would gain some feedback.

1. After changing the timestamp format, only new messages currently get the new format. Either we would have to refresh the page or reformat everything (not spent much time investigating how that would be done, but I imagine it would mean essentially clearing all messages out of the dom first)
2. The spacing is a little weird just now. With the specified width, it wouldn't work with longer timestamps (it just overlaps the name), but with the way I've done it, the spacing is different for most rooms.

With HH:mm timestamp
screenshot from 2016-01-22 22-25-28

After changing to HH:mm:ss timestamp:
screenshot from 2016-01-22 22-26-28

After refreshing:
screenshot from 2016-01-22 22-26-20


client/js/shout.js
```diff
@@ -382,6 +382,28 @@ $(function() {
 		}
 	}
 
+	if ($.cookie("timestamp")) {
```
xPaw: Use the existing `settings` cookie object, it will be easier to switch to another storage method in the future.

xPaw:
Regarding the non changing timestamps: `.time` could have a data attribute with a raw timestamp, then when the setting is changed, you can do `chat.find('.time')` and reformat it easily.

client/js/shout.js
```diff
@@ -382,6 +382,28 @@ $(function() {
...
+			"tz", function(time) {
+				if (time) {
+					var utc = moment.utc(time, timestamp).toDate();
+					return moment(utc).format(timestamp);
+				} else {
+					return "";
+				}
+			}
+		);
+		$.cookie("timestamp", timestamp);
```
xPaw: When cookie is loaded on lines 385-389 it calls `change` event, and then the change event changes the cookie again.

client/js/shout.js
```diff
@@ -382,6 +382,28 @@ $(function() {
...
 
+	if ($.cookie("timestamp")) {
+		$("#timestamp").val($.cookie("timestamp")).change();
+	} else {
+		$("#timestamp").val("HH:mm").change();
+	}
+
+	$("#timestamp").on("change", function() {
+		var timestamp = $(this).val();
+		Handlebars.unregisterHelper("tz");
```
xPaw: Is there really a need to re-register the handler? Can't you just make `timestamp` variable global?

YaManicKill: Well, there are a couple of issues with that:

1. I don't like global variables, they get quite messy. If the handler was in a module that we passed the variable to, then I'd be all for that, but the way it works, that's not great.
2. The module that the handler is in currently minified, which isn't a big issue, but does mean I'd need to set it to window.timestamp rather than just timestamp as well.

Basically, with the way we currently have the code, I think this is probably the least of the evils.

client/js/shout.js
YaManicKill: Yeah, it's alright, I'm changing how this is done anyway, I'm modifying the generic cookie setting.

client/js/shout.js
xPaw: Maybe I didn't phrase it correctly, but I didn't mean global as in `window`. But rather the parent object which both the settings and helper functions can access.

YaManicKill: Ah wait, yes, of course. I can do that, but I would still need to have the helper register be out of it's existing file, which isn't ideal, as that's less modularising rather than more :-P

YaManicKill:
Alright, I have reverted my css change, because it was just not ideal. I'll have a think later about a better way to do it.

I've also changed the time handler to use a variable rather than re-rendering it. It's not ideal, but it's better than the other options, I think.

Finally, I've modified the cookie to use the same cookie and the same setting/desetting code.

Still to do:

1. Sort out the css for the time
2. Find a good way to rerender times (I'll look into @xPaw 's idea)

client/js/libs/handlebars/tz.js
```diff
@@ -1,8 +1,10 @@
+var timestamp = window.timestamp;
```
xPaw: I'd suggest keeping the variable instead of the shout.js's anon function scope. But if you're keeping it in `window`, then you should give it a better name (ideally with a namespace).

client/js/libs/handlebars/tz.js
```diff
@@ -1,8 +1,10 @@
...
+
 Handlebars.registerHelper(
 	"tz", function(time) {
 		if (time) {
-			var utc = moment.utc(time, "HH:mm:ss").toDate();
-			return moment(utc).format("HH:mm");
+			var utc = moment.utc(time, timestamp).toDate();
```
xPaw: Changing "HH:mm:ss" to variable shouldn't be done here probably? As that's what the server sends.

YaManicKill: Ah right, that explains why they were different. The server should probably send the whole datestamp, otherwise there are only hour, minutes and seconds possible, and this becomes pointless :-P

I'll take a look at sending the whole thing back from the server.

client/js/libs/handlebars/tz.js
YaManicKill: If it isn't kept in window, it can't be used here at all (because of minification). When we can properly modularise the code (and pass state between them) then it'll be much better. For just now, I think this is the best solution.

But yeah, of course, that's a good point. It's still early in the morning :-P

xPaw: It can be used like that, if you register the handler inside of shout.js too.

YaManicKill: I'm not sure I understand what you are suggesting?

client/js/shout.js
```diff
@@ -374,18 +374,34 @@ $(function() {
...
 		quit: true,
 		notifyAllMessages: false,
+		timestamp: "HH:mm"
 	}, $.cookie("settings"));
 
+	$("#timestamp").on("change", function() {
+		var self = $(this);
+		window.timestamp = self.val();
+		$(".time").each(function() {
+			var newtime = window.renderTime(self.attr("data-time"));
```
xPaw: You can directly invoke `Handlebars.helpers.tz`, no need to make a new window function.

YaManicKill: Oh, awesome. Didn't realise that.

src/models/msg.js
```diff
@@ -27,7 +27,7 @@ function Msg(attr) {
 		from: "",
 		id: id++,
 		text: "",
-		time: moment().utc().format("HH:mm:ss"),
```
YaManicKill: This is removed so that we send the whole timestamp to the client, so people can format it how they like.

client/views/msg.tpl
```diff
@@ -1,6 +1,6 @@
 {{#each messages}}
 <div class="msg {{type}} {{#if self}}self{{/if}}">
-	<span class="time">
+	<span class="time" data-time="{{time}}">
```
YaManicKill: Easy place to keep the full timestamp so we can re-render later on if they change the format.

client/js/shout.js
```diff
@@ -374,18 +374,34 @@ $(function() {
...
+		} else if (options[i] !== false) {
+			settings.find("input[name=" + i + "]").val(options[i]);
 		}
 	}
 
 	settings.on("change", "input", function() {
 		var self = $(this);
 		var name = self.attr("name");
-		options[name] = self.prop("checked");
+		if (self.is(":checkbox")) {
```
YaManicKill: Changed so we can have more than just checkboxes use the default settings here.

client/js/shout.js
```diff
@@ -374,18 +374,34 @@ $(function() {
...
+			self.html(newtime);
+		});
+	});
+
 	for (var i in options) {
-		if (options[i]) {
+		if (options[i] === true) {
 			settings.find("input[name=" + i + "]").prop("checked", true);
+		} else if (options[i] !== false) {
+			settings.find("input[name=" + i + "]").val(options[i]);
```
YaManicKill: Also for non-checkbox types.

client/js/shout.js
```diff
@@ -374,18 +374,34 @@ $(function() {
 		thumbnails: true,
 		quit: true,
 		notifyAllMessages: false,
+		timestamp: "HH:mm"
 	}, $.cookie("settings"));
 
+	$("#timestamp").on("change", function() {
+		var self = $(this);
+		window.timestamp = self.val();
+		$(".time").each(function() {
```
YaManicKill: Re-renders all existing timestamps.

xPaw:
Here's an interesting thought: if the timestamp setting is empty, it shouldn't invoke `moment` to format the time, and `.time` element would be hidden.

YaManicKill:
Yeah, @xPaw, not to sound like I'm just stealing your idea but I was actually thinking about that today. I'm not sure how user-friendly that is, but it would mean 1 less setting being added in the future (to turn off timestamps).

Anyone else got thoughts on that?

xPaw:
> I'm not sure how user-friendly that is, but it would mean 1 less setting being added in the future

Adding a description next to the input would help.

astorije:
Hey @YaManicKill, I haven't reviewed the code of this yet, but I wanted to share my two cents.

*To give you a few words of context, the last projects I have been involved in (mostly for work) either focused on or left out entirely usability and user experience, and that made a whole world of difference. Since then, I have been very careful about the UX of products I use or develop, and since my company recently hired a dedicated UX lead, I can't promise I won't get even more pedantic about it :D*

I am really not fond of having a timestamp format input at all. This is one of those cases where customization complexifies things unnecessarily. Also, an empty input meaning no timestamp is very cryptic (even with a help description, which shouldn't necessary for this simple setting).

If you think about it, the 2 cases `HH:mm` and `HH:mm:ss` cover pretty much everything users will set. Add to that 24-hour and 24-hour formats (to not leave our American friends :-) ) and really that's about it. I don't possibly see a use case where only hours, minutes, or seconds would be relevant to more than 0.001% of our users :-)
(And if really someone wants to have such a customized setup, then that's where the package system comes in handy!)

Just to see how design hipsters are doing, I opened the time settings of the Mac I'm on right now:

screen shot 2016-01-25 at 23 54 05

This is very simple and clear IMO. If we relate to that, one checkbox for displaying seconds, one checkbox (or 2 radio buttons / a select element) for the 12/24-hour format switch, and deal! :-)
I prefer 2 simple, descriptive checkboxes, or even 4, to a fully customizable input field that requires a description and maybe a link to a doc (to detail `HH`, `mm`, ...) to be used.

Do these arguments make sense to you? I'm sorry to receive your PR negatively while you have been contributing a lot to this project though :-/

Internally, keeping a string for the time format makes a lot of sense though (I mean, even if we didn't need one to give to moment.js), as it offers a possibility to extend the system by manipulating it.

astorije:
Mentioning https://github.com/erming/shout/pull/484 here for the record.

YaManicKill:
Fair point.

YaManicKill:
Ok, to clean up, I'll remove this one, and we can look at just redoing #578 and #484 (with possibly a second one for mobile).

PR #609, opened 2016-01-24
Simply parse all 0-99 irc colors

xPaw:
This does a couple of things:

- Removes the unnecessary `colourMap` array
- Colors in invalid range (16-99) are ignored, even though the class is inserted into css (to allow for future style if needed), they remain uncolored


astorije:
Not even a chance to look at it, nice.

client/js/libs/handlebars/parse.js
```diff
@@ -56,8 +56,8 @@ function uri(text) {
  * https://github.com/megawac/irc-style-parser/tree/shout
  */
 var styleCheck_Re = /[\x00-\x1F]/,
-    back_re = /^(\d{1,2})(,(\d{1,2}))?/,
-    colourKey = "\x03", colour_re = /\x03/g,
+    back_re = /^([0-9]{1,2})(,([0-9]{1,2}))?/,
```
astorije: Why replacing `\d` with `[0-9]`? Aren't they the same?

xPaw: They are not. http://stackoverflow.com/a/6479605/2200891

astorije: Waouh, I'd pretend I didn't know that, but I also see I +1ed this answer more than 2 years ago :-)

client/js/libs/handlebars/parse.js
```diff
@@ -108,20 +99,19 @@ function colors(line) {
...
-        // set the background colour
-        // we don't overide the background local var to support nesting
-        if (colourMap[+match[3]]) {
-            background = " " + colourMap[+match[3]].back;
-        }
+		colour = "irc-fg" + +match[1];
+		// set the background colour
+		if (match[3]) {
+			background = " irc-bg" + +match[3];
+		}
```
astorije: Bad indendation.

xPaw: This whole file needs to be converted to tabs, and eslint ignores `handlerbars/` folder.

PR #611, opened 2016-01-24
Update dependancies

xPaw:
I hope I didn't miss any breaking changes. Couldn't update `socket.io` as it kept complaining about parse error.


package.json
```diff
@@ -28,25 +28,25 @@
...
-    "cheerio": "^0.17.0",
+    "cheerio": "~0.19.0",
     "commander": "^2.3.0",
     "event-stream": "^3.1.7",
     "express": "^4.9.5",
-    "lodash": "~2.4.1",
+    "lodash": "~4.0.0",
     "mkdirp": "^0.5.0",
-    "moment": "~2.7.0",
+    "moment": "~2.11.1",
```
astorije: For the dependencies you updated, let's use the non tilde-range (`~`) version: `cheerio` in `0.19.0` instead of `~0.19.0`, ...

Having tilde-ranges gives you the false impression you stay latest why this cannot be ensured. I've noticed that projects reaching a decent size switch to the fixed versions as well. Let's do that as we upgrade dependencies :-)

astorije:
Hey @xPaw, thanks for this! Don't you think we should have one PR per upgraded prod dependency (and an additional one for all `devDependencies`)?
I believe it would be easier to review, to test and to potentially debug :-)

YaManicKill:
@astorije I would agree for breaking changes. I think it's fine to keep everything else in the same one, but let's keep lodash in a different PR as it's such a big jump. We can test it for a bit and make sure nothing else has gone wrong. I might have time sometime later this week to also read through the breaking changes.

@xPaw Did you remember to check both the 3.x **and** the 4.x breaking changes for lodash?

xPaw:
@YaManicKill I did, yes.

astorije:
I agree with @YaManicKill. @xPaw, do you mind opening a PR just for lodash? :)

xPaw:
This PR now only upgrades cheerio and moment, and drops `^` `~` from versions

williamboman:
> drops ^ ~ from versions

Wouldn't it be better to bump versions to the latest version within the current semver range as well? This PR effectively downgrades most deps.

astorije:
@williamboman, the transition is awkward for sure, but it's the safest way to go on long-term.

@xPaw, what about removing `~`/`^`-ranges as we upgrade the versions, and keep these to the packages we haven't upgraded yet, unless this is already the latest version? I feel it's a good compromise.

xPaw:
You're right about versions, I've upgraded them. Left out lodash as there's a separate PR, and socket.io because it breaks, and the client lib needs updating too.

YaManicKill:
Oh? What breaks about socket.io? I did a quick upgrade of it sometime past and didn't notice anything, but I didn't look at everything, and didn't look at the release notes for breaking changes.

I'll add a ticket for that specifically, because it'd be good to upgrade that as there are a number of issues with the version we are currently on.

xPaw:
The page kept reloading for me due to a decoding error.

MaxLeiter:
Discussion on updating client libs is here just in case people cant find it: https://github.com/erming/shout/issues/562

xPaw:
Closing in favor of https://github.com/thelounge/lounge/pull/8

PR #613, opened 2016-01-24
Initial work on context menus

xPaw:
chrome_2016-01-25_15-47-56

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -982,6 +982,62 @@ button, .user {
 	width: 58px;
 }
 
+#context-menu-container {
+    display: none;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    z-index: 1000;
+    background: transparent;
+}
+#context-menu {
+    position: absolute;
+    list-style: none;
+    margin: 0;
+    padding: 4px 0;
+    width: 240px;
+    background-color: #fff;
+    border: solid 1px #dfdfdf;
+	box-shadow: 0 0 20px 0 rgba(0,0,0,.2);
+}
+.context-menu-item {
+    display: block;
+    margin-bottom: 4px;
+}
+.context-menu-item:last-child {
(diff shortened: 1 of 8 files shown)


MaxLeiter:
Just adding my two cents: I think on mobile the context menu should exist elsewhere as a button, similar to the 'leave' one. Perhaps we can replace the current leave channel ('x') button with a dropdown arrow?

xPaw:
I would rather try to make long press work instead of adding arrows.

YaManicKill:
I agree with @xPaw 

For a button to work, we would need to increase the size of the sidebar, as trying to hit a button on that small name is near impossible currently (it's hard enough to hit the expando for link expander just now as it is).

MaxLeiter:
If long press opens a context menu then how would mobile users rearrange channels and servers?

xPaw:
At least in Chrome on Android long press simply selects the text, you can't rearrange servers.

xPaw:
Figured out how to make the context menu work on mobiles, by adding `user-select: none;` on the sidebar.

photo_2016-01-30_23-13-13

YaManicKill:
I really like this so far, looks really nice and it brings forward many features to the user, especailyl on mobile.

I think having a longpress on mobile, and right-click on desktop works really nicely as these are things people expect to do something like that.

MaxLeiter:
I agree that longpress make sense on mobile, but I think there should be a way to reorganize channels and servers. That could be built into the context menu (ie “unlock” or “move down”) but that seems like an awkward UX choice

YaManicKill:
We could have a "grabber" (I forget what the proper word for them is) on touch-screen devices that you can use for reordering.

xPaw:
The question is why current grabbing stuff does't work on mobile. That should be triggered long before the contextmenu event.

xPaw:
Closing this in favor of https://github.com/thelounge/lounge/pull/9

Issue #577, opened 2016-01-11
Display actions (such as join/leave/nick change) differently


xPaw:
It's confusing to display them same as normal messages.

Here's how it's done in HexChat for example:
![hexchat_xpaw_ _holmes_freenode__ steamdb_2016-01-11_13-39-21](https://cloud.githubusercontent.com/assets/613331/12232623/c48e2e4e-b868-11e5-849d-4ed9bf304ff5.png)

astorije:
For me (default theme), the "join" and "quit" messages are displayed in a very light gray, so not like normal messages. Maybe we could make this even better?
Feel free to suggest things here and if we can reach to a simple consensus, maybe you could open a PR :P

Thanks for the report!

dgw:
The main issue I see with the presentation of join/part/quit/nick events is that they follow an identical line format. Being light gray just makes them hard to read; they still look like talking.

For example, here is how my theme for Textual displays various events:
![image](https://cloud.githubusercontent.com/assets/164140/12286731/e44c96f8-b994-11e5-91c4-e3b71b6cf101.png)

The non-speech events are all a different size, color, and line layout.

xPaw:
Yeah, simply changing color is not enough. @dgw's Textual screenshot is how it should be implemented.

PR #617, opened 2016-01-27
Rotate part/quit icon

MaxLeiter:
**Before:**

image

**After:**

image

Before, the icons were aiming the same direction. That threw me off and I feel that this looks better (and better represents leaving)

### DIFF
--- a/client/css/style.css
+++ b/client/css/style.css
@@ -678,9 +678,11 @@ button, .user {
 }
 #chat .part .from:before,
 #chat .quit .from:before {
-    font-family: FontAwesome;
+	font-family: FontAwesome;
 	content: "\f08b";
 	color: #FF4136;
+	display: inline-block;
+	transform: rotate(180deg);
 }
 #chat .topic .from:before {
     font-family: FontAwesome;


astorije:
Yes! Big :+1: on that! This is among the list of things that I noticed could be improved from #577 (so many thanks to @xPaw!). It does look much better indeed!

Just one thing, could you squash these into 1 commit and keep the "Rotate part/quit icon" message please? (Also, small detail, if you look at the CONTRIBUTING file and best practices in Git, always phrase your commit message with infinitive tense)

@xPaw, just curious, are you OK with this? :-)

Thanks @MaxLeiter!

MaxLeiter:
Bump

client/css/style.css
```diff
@@ -678,9 +678,15 @@ button, .user {
 }
 #chat .part .from:before,
 #chat .quit .from:before {
-    font-family: FontAwesome;
+	font-family: FontAwesome;
 	content: "\f08b";
 	color: #FF4136;
+	display: inline-block;
```
astorije: What happens if you don't specify this?

MaxLeiter: Inline elements can't be transformed, and pseudo selectors are automatically inline.

http://stackoverflow.com/questions/9779919/css-rotate-a-pseudo-after-or-before-content

astorije: Great, I'll sleep smarter!

client/css/style.css
```diff
@@ -678,9 +678,15 @@ button, .user {
...
+	-webkit-transform: rotate(180deg);
+	-moz-transform: rotate(180deg);
+	-ms-transform: rotate(180deg);
+	-o-transform: rotate(180deg);
+	transform: rotate(180deg);
```
xPaw: I would drop all prefixes. http://caniuse.com/#search=transform

JocelynDelalande: That would be more consistent with what we use in other parts of the CSS (that would break anyway with IE<10). 

BTW, it's not to be done here, but the -webkit-transform hanging around in the CSS should IMHO be removed, if there is a "special needs" browser which requires to use prefixed property for transform, it's definitely not chrome (see caniuse).

JocelynDelalande:
Ok, reviewed, makes sense. Once CSS vendor-prefixes are dropped as suggested by @xPaw , you get my auto-:+1: :-)

client/css/style.css
astorije: I was thinking about that this morning actually, glad someone stepped up :)
Completely safe to remove the prefix-vendor versions here.
The spec has been stagnant for more than 2 years (in WD state, which is a bit worrying actually), so browsers have definitely had time to catch up with this. Plus, s***w old browsers :-)

YaManicKill: Plus, if we really care about vendor prefixes, we should use auto prefixer.

astorije: (Let's not do that, this is a fairly modern web app, it's reasonable to ask users to have recent browsers, and we can start ensuring compatibility when the project is more stable.)

YaManicKill: Oh, I'm not suggesting we should do it just now, I'm just saying that if we were to do it (which is a whole different discussion) then we would use auto-prefixer, not do it manually.

MaxLeiter:
Bump :)

astorije:
@MaxLeiter, squashy squash? :) I'll merge afterwards.

Issue #604, opened 2016-01-23
Use localStorage instead of cookies for storage


xPaw:
Using cookies is silly to persist settings. Ideally logged in users would persist their settings on the server.

MaxLeiter:
Alternaitvley, we can update to [js-cookie](https://github.com/js-cookie/js-cookie/tree/v1.5.1), jquery.cookie.js's successor (what we currently use).

xPaw:
That doesn't change the fact that settings are stored in cookies, which is not a good idea.

astorije:
I agree, we'd better off using localStorage, and that is also highlighted by Shout becoming a more and more powerful web app.

I do believe however that this is just one step among the grand scheme of clean user settings management. I started to gather broad thoughts on design/requirements of user settings and will publish them soon, probably in a wiki page or an issue, so that we can discuss about it and reach the best solution :-)
It would be interesting to also gather up all opened tickets (and maybe closed too?) that are related to that as they add or specify requirements.

MaxLeiter:
It could be beneficial to save settings server-side if shout is in private, and then use localStorage if public. That way, a users settings can persist across devices. Users already have a config file for saving channels and networks

xPaw:
Fixed in lounge.

PR #623, opened 2016-01-27
Update chat template so no users-icon shown when a server or private conversation is selected.

### LINKED ISSUE #622: Remove User List icon when viewing Server Connection
1. Click the Server Connection on the left pane
2. Look in the right hand upper corner

Observed:
There is a User List icon, that does nothing when you click it.

Expected:
That shouldn't be there. Better would be the settings (gear) icon for that server.

MaxLeiter:
Fixes #622, should I add a font-awesome settings icon even if it won't do anything at the moment?

### DIFF
--- a/client/views/chat.tpl
+++ b/client/views/chat.tpl
@@ -2,7 +2,9 @@
 <div id="chan-{{id}}" data-title="{{name}}" data-id="{{id}}" data-type="{{type}}" class="chan {{type}}">
 	<div class="header">
 		<button class="lt"></button>
-		<button class="rt"></button>
+		{{#equal type "channel"}}
+			<button class="rt"></button>
+		{{/equal}}
 		<div class="right">
 			<button class="button close">
 				{{#equal type "lobby"}}
--- a/client/js/shout.templates.js
+++ b/client/js/shout.templates.js
@@ -37,8 +37,11 @@ templates['chat'] = template({"1":function(depth0,helpers,partials,data) {
     + escapeExpression(((helper = (helper = helpers.type || (depth0 != null ? depth0.type : depth0)) != null ? helper : helperMissing),(typeof helper === functionType ? helper.call(depth0, {"name":"type","hash":{},"data":data}) : helper)))
     + "\" class=\"chan "
     + escapeExpression(((helper = (helper = helpers.type || (depth0 != null ? depth0.type : depth0)) != null ? helper : helperMissing),(typeof helper === functionType ? helper.call(depth0, {"name":"type","hash":{},"data":data}) : helper)))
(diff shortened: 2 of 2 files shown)


xPaw:
:+1: 

> should I add a font-awesome settings icon even if it won't do anything at the moment?

Nah. Channels eventually should have settings too.

YaManicKill:
Definitely don't add a settings icon yet.

However, yeah, I don't think you should set this up as if the settings is only going to be on the lobby. I would just show user on anything other than lobby or user.

MaxLeiter:
Thats how it works with his PR @YaManicKill

astorije:
> should I add a font-awesome settings icon even if it won't do anything at the moment?

Definitely not, or we'll have a lot of "Settings don't work" issues popping up :-)

This is an off-topic side note that we should keep for later (the user settings design doc I was mentioning in #604), but I'm concerned about UI/UX regarding having a Settings button at the bottom left of the app (for global settings) and on the top right of each channel (for channel settings). I have ideas about that and will share them.

client/views/chat.tpl
```diff
@@ -2,7 +2,11 @@
 <div id="chan-{{id}}" data-title="{{name}}" data-id="{{id}}" data-type="{{type}}" class="chan {{type}}">
 	<div class="header">
 		<button class="lt"></button>
-		<button class="rt"></button>
+		{{#equal type "lobby"}}
+			{{!-- Insert settings icon --}}
+		{{else}}
+			<button class="rt"></button>
+		{{/equal}}
```
astorije: I'm very not fond of placeholders like this one for potential future code/UI elements that might not be located there. Could you instead check that the type is _not_ lobby then display the button, without `else` statement please?

astorije:
By the way, thanks a lot for that!! :-)

client/views/chat.tpl
astorije: Actually, checking that the type is a channel is probably more appropriate to make sure it doesn't show up on lobbies and private messages. Anything else we should display or not display this?

xPaw: You're right, changing it to `{{#equal type "channel"}}` will do the job.

astorije:
@MaxLeiter, just to confirm, is `shout.templates.js` up-to-date with your change on `chat.tpl`?

MaxLeiter:
I forgot to run it a second time; will update tonight and message in IRC when I do.

MaxLeiter:
@astorije looks like I did update the template; ready to pull :)

astorije:
:+1:

MaxLeiter:
Worth noting that checking if its a channel also solved the issue of a user icon showing in private conversations.

PR #627, opened 2016-01-29
No longer notify on messages sent by yourself from different tab. Fixes #592

### LINKED ISSUE #592: Multiple tabs open cause notifications of own messages
When you have multiple tabs with Shout open, you get notifications for own messages you write. For example in a private conversation.

Also, you get notifications for the whois info which is displayed when starting a private conversation.

Both things should not happen by checking for that since it doesn’t make sense to send you notifications for that since you either just saw that / just sent that.

dubzi:
Messages from yourself, or whois responses, no longer trigger a notification on a second client you may have running.


client/js/shout.js
```diff
@@ -644,10 +644,12 @@ $(function() {
 		var button = sidebar.find(".chan[data-target=" + target + "]");
 		var isQuery = button.hasClass("query");
 		var type = msg.type;
+		var self = msg.self;
 		var highlight = type.contains("highlight");
 		var message = type.contains("message");
+		var whois = type.contains("whois");
 		var settings = $.cookie("settings") || {};
-		if (highlight || isQuery || (settings.notifyAllMessages && message)) {
+		if ((highlight || isQuery || (settings.notifyAllMessages && message)) && !self && !whois) {
```
xPaw: Why not move this code down after checking `ignore` array so that the same filtering applies?

dubzi: True, I could probably change that part a bit since it's now doing some checks twice (such as whether or not the channel receiving the message is active or not). Once for the notification, and once for the badge.

dubzi:
Integrated xPaw's comment in this latest commit. As a result, you'll now also no longer receive a badge next to a channel if you were the one sending the message using a second client.

client/js/shout.js
```diff
@@ -642,37 +642,13 @@ $(function() {
...
-						button.click();
-						this.close();
-					};
-					window.setTimeout(function() {
-						notify.close();
-					}, 5 * 1000);
-				}
-			}
+
+		if (msg.self) {
```
xPaw: You can have this `if` the first thing in the function.

client/js/shout.js
```diff
@@ -682,18 +658,47 @@ $(function() {
...
-		if (badge.length !== 0) {
-			var i = (badge.data("count") || 0) + 1;
-			badge.data("count", i);
-			badge.html(i > 999 ? (i / 1000).toFixed(1) + "k" : i);
-			if (highlight || isQuery) {
-				badge.addClass("highlight");
+		var isQuery = button.hasClass("query");
+		var highlight = type.contains("highlight");
+		var message = type.contains("message");
+		var settings = $.cookie("settings") || {};
```
xPaw: Can't you access `options` variable directly here? I know I'm nitpicking on the existing code, but might as well improve it.

client/js/shout.js
```diff
@@ -682,18 +658,47 @@ $(function() {
...
+					button.click();
+					this.close();
+				};
+				window.setTimeout(function() {
+					notify.close();
+				}, 5 * 1000);
+			}
+		}
+
+		if (!button.hasClass("active")) {
```
xPaw: On line 651 you have the same conditional, look into that.

client/js/shout.js
dubzi: No worries, I appreciate the feedback. I wasn't familiar with the `options` variable, but you're right, i'll use that instead.

client/js/shout.js
dubzi: They're different though. The first condition checks if the channel is active and the window is active. If they are, we're not showing notifications or badges. The second condition only checks if the channel is active. If it's active, the badge of unread messages isn't displayed.

xPaw: Ah yeah, that makes sense. Just cache `hasClass` result into a variable, so you're not calling it twice.

xPaw:
:+1: looks good

astorije:
FYI, this is under review for The Lounge: https://github.com/thelounge/lounge/pull/6

PR #631, opened 2016-02-01
Handle invite event

### LINKED ISSUE #625: Invite messages don't show channel
"X invited Y into the channel" <-- you get this in the lobby, and it doesn't say what channel.

It would be good to either show the message in the channel itself, or to say

"X invited Y into the channel (#shout-irc)"

xPaw:
Fixes #625

### DIFF
--- a/client/js/shout.js
+++ b/client/js/shout.js
@@ -193,8 +193,10 @@ $(function() {
 
 		var chan = chat.find(target);
 		var from = data.msg.from;
+		var msg;
 
 		if ([
+			"invite",
 			"join",
 			"mode",
 			"kick",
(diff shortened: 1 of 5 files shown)


client/js/shout.js
```diff
@@ -205,6 +207,7 @@ $(function() {
 			"action",
 		].indexOf(type) !== -1) {
 			switch (type) {
+			case "invite": data.msg.formattedAction = "invited you to"; break;
```
YaManicKill: So, is this just for invites that someone sends to you. What I was referring to was invites that are sent to someone else. Will this be both of them? If so, the wording doesn't work for all of them.

xPaw: IRC protocol does not support showing when someone invites to the channel.

However there is http://ircv3.net/specs/extensions/invite-notify-3.2.html

MaxLeiter:
For those curious, this is the FA icon used for invites: http://fortawesome.github.io/Font-Awesome/icon/envelope/

erming:
Unfortunately I got this:

image

Running debug, the IRC message looks like this:
```
INVITE :erming!~erming@46.101.57.23 INVITE test-invites :#flerp
```

Maybe it works just as fine with `<name> invited you to #chan`?

xPaw:
@erming should be fixed.

astorije:
Hey @xPaw, would you mind copying this PR over at https://github.com/thelounge/lounge please?
Thanks!

Issue #628, opened 2016-01-30
Strip color codes from notifications


xPaw:
Otherwise you end up with this:
![2016-01-30_22-43-18](https://cloud.githubusercontent.com/assets/613331/12698337/f8f200fe-c7a2-11e5-85bf-74bea0ed7b69.png)

xPaw:
Fixed in lounge.

PR #634, opened 2016-02-01
Strip color codes from notifications

sprusr:
Fixes erming/shout#628


dgw:
This should probably strip all control codes, not just colors.

- `\x02`: Bold
- `\x03`: colored text (already handled with arguments, :+1:)
- `\x1D`: italic text
- `\x1F`: underlined text
- `\x16`: swap background and foreground colors ("reverse video")
- `\x0F`: reset all formatting

sprusr:
Hey @dgw thanks for the suggestion. Initially I thought it wouldn't be necessary to handle these, as they represent characters that won't be displayed, but I suppose it depends on the situation.
Will make the changes suggested, thanks again.

dgw:
If only worrying about characters that aren't displayed, there's no reason to remove `\x03` from colors; just removing the digits would suffice. :wink:

Thanks for updating the PR. I figured (and should have explained initially) that it's best to remove all of the control codes in case of misbehaving fonts, or other weirdness that could arise from having non-printing characters in the notifications. There are a lot of system configurations out there, and as long as we're stripping _one_ kind of control character…

:100::+1:

xPaw:
:+1: However, this probably conflicts #627 as that PR moves code around.

client/js/shout.js
```diff
@@ -655,7 +655,7 @@ $(function() {
 				favico.badge("!");
 				if (settings.badge && Notification.permission === "granted") {
 					var notify = new Notification(msg.from + " says:", {
-						body: msg.text.trim(),
+						body: msg.text.trim().replace(/\x02|\x1D|\x1F|\x16|\x0F|\x03(?:\d{1,2}(?:,\d{1,2})?)?/, ""),
```
xPaw: Perhaps it should `trim` after `replace`?

sprusr: Yeah, thanks for noticing that. Will make suggested change.

xPaw: While you're at it, change `\d` to `[0-9]`. See http://stackoverflow.com/a/6479605/2200891

sprusr:
My apologies for the number of commits this is taking for a single regex! I haven't made much contribution to open source projects before, so thank you for being so helpful along the way.

YaManicKill:
Don't worry about that, @sprusr, we all have to start somewhere :-) 

The only thing is that before this gets merged, it'll be good to squash the commits into 1.

PR #639, opened 2016-02-04
Fix comments in config file

FryDay:
Some of the defaults in the comments do not match the actual default values.


MaxLeiter:
:+1:, this just came up in IRC (someone pasted their shout config and some information was off, confused me.)

PR #640, opened 2016-02-04
Disable autocomplete on the input field

ralt:
This avoids having the history of what was written in the field displayed every time you click on it.


dgw:
Only need this OR #567, depending on which is the "better" (more reliable) solution.

PR #643, opened 2016-02-07
package.json: lint all .js files

williamboman:
This makes the linting process less prone to miss new (or even
existing) files. Ignoring specific file paths is delegated to the
.eslintignore file in the root directory.

I just noticed this, not sure why I had explicitly set files to lint to begin with..

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "start": "node index",
     "build": "grunt",
     "test": "HOME=test/fixtures mocha test/**/*.js && npm run lint",
-    "lint": "eslint index.js Gruntfile.js src/ test/ client/ defaults/"
+    "lint": "eslint ."
   },
   "keywords": [
     "browser",


YaManicKill:
Are there no new errors thrown up? Because this will lint the libraries.

williamboman:
> Because this will lint the libraries.

Which libraries?

MaxLeiter:
@williamboman /node_modules/*

williamboman:
@MaxLeiter ESLint skips `node_modules` and `bower_components` by default, I don't even think you can override that. Like, at all :P.

MaxLeiter:
then I believe @YaManicKill meant everything in /client/libs, but I'm not sure if those were being lint before or not. If /client/libs will now be lint and aren't currently, I think this PR should wait until client-side dependencies are on some kind of package manager (#562)

williamboman:
> then I believe @YaManicKill meant everything in /client/libs, but I'm not sure if those were being lint before or not.

Ah, no, they are currently [not being linted](https://github.com/erming/shout/blob/master/.eslintignore).

YaManicKill:
Yeah, I means client/libs

Didn't notice you had them ignored already. Looks good then, it shouldn't change anything for now, I think, but it's a good change to make.

MaxLeiter:
:+1: from me

astorije:
@williamboman, would you mind opening this to [The Lounge](https://github.com/thelounge/lounge) fork please? Thanks!

williamboman:
@astorije Done (https://github.com/thelounge/lounge/pull/42)

PR #644, opened 2016-02-08
add sasl support

flyinprogrammer:
This only works when https://github.com/slate/slate-irc/pull/37 is pulled in.
You can edit the package.json if you want ?
Or wait for those merges.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -158,13 +158,31 @@ Client.prototype.connect = function(args) {
 	var irc = slate(stream);
 	identd.hook(stream, username);
 
-	if (args.password) {
-		irc.pass(args.password);
-	}
+
 
 	irc.me = nick;
-	irc.nick(nick);
-	irc.user(username, realname);
+	if (args.sasl) {
+		irc.cap_req("sasl");
+		irc.nick(nick);
+		irc.user(username, realname);
+		irc.once("cap", function() {
+			irc.authenticate("PLAIN");
+			irc.once("authenticate", function() {
+				irc.authenticate64(username, args.password);
+				irc.once("saslsuccess", function() {
+					irc.cap_end();
+				});
+			});
+		});
+	} else {
+		if (args.password) {
+			irc.pass(args.password);
+		}
+		irc.nick(nick);
+		irc.user(username, realname);
+	}
+
+
 
 	var network = new Network({
 		name: server.name,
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,2 +1,77 @@
-node_modules/
-npm-debug.log
+### Node template
```
YaManicKill: We really don't need all this stuff in the gitignore. Was there something in particular that was causing you issues and wanting to be added to the repo?

src/client.js
```diff
@@ -158,13 +158,31 @@ Client.prototype.connect = function(args) {
...
 
-	if (args.password) {
-		irc.pass(args.password);
-	}
+
 
 	irc.me = nick;
-	irc.nick(nick);
-	irc.user(username, realname);
+	if (!!args.sasl) {
```
YaManicKill: Just do if(args.sasl)

Never use "!!" it's very confusing for people who don't know what it means (and there's no real advantage to it. In this case, args.sasl should only ever be a boolean, so theres no need to turn it into a boolean.

Xe: With how javascript implicitly typecasts things, it is probably better from a realistic computer science point of view to do this.

flyinprogrammer: i am just a script kitty: https://github.com/flyinprogrammer/shout/blob/just_sasl/src/client.js#L191

.gitignore
flyinprogrammer: i just use intellij, they have a plugin that generates from templates, this is the JetBrains and Node template.

williamboman: @flyinprogrammer See https://github.com/slate/slate-irc/pull/37#issuecomment-181505599

flyinprogrammer:
changed to 'args.sasl' and dropped my .gitignore

deviousgeek:
You da man.
`(σ・・)σ :: `

MaxLeiter:
:+1: looks good to me

svbeon:
Pull this in, now

flyinprogrammer:
if you want this code today, it can be used via docker like this:

`docker run -dP flyinprogrammmer/shout:with_flyinprogrammer_slate-irc`

[Dockerfile](https://hub.docker.com/r/flyinprogrammer/shout/~/dockerfile/)

svbeon:
Hmm, it's a bit unclear here where to put the sasl password, maybe change the password box text to "Password/SASL password"

PR #649, opened 2016-02-11
Add dummy method to forms (quiets LastPass)

### LINKED ISSUE #648: LastPass "LastPass detected a login form that is insecure"
When I log in I get the error "LastPass detected a login form that is insecure" from LastPass. I am running my instance behind a proxy which add https. I'm not sure why LastPass is saying this, but I did find this post. https://lastpass.com/support.php?cmd=showfaq&id=7336

I did not enable https in shout because of the proxy providing it.

maxpoulin64:
It appears LastPass looks for the use of the GET method in forms on the page to determine if it is secure or not. Since Shout never actually submits any of the forms, no method is set, which means the default of GET is used. Changes form methods to POST to quiet the warning. Should fix #648

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -46,7 +46,7 @@
 					<div class="header">
 						<button class="lt"></button>
 					</div>
-					<form class="container" action="">
+					<form class="container" method="post" action="">
 						<div class="row">
 							<div class="col-xs-12">
 								<h1 class="title">Sign in</h1>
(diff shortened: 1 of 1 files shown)


YaManicKill:
Huh, that seems like a very brittle way of figuring things out from lastpass's view of things. But whatever. Yeah, seems like a good fix.

astorije:
FYI, this has been moved over to The Lounge: https://github.com/thelounge/lounge/pull/19
Should we close this one?

PR #663, opened 2016-05-10
Fix: colored nicknames for new info messages design

bews:
Now they're colored too (if colored nicknames option is enabled).

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,15 +28,9 @@ shout --help
 
 For more information, read the [documentation](http://shout-irc.com/docs/).
 
-## Development setup
+## Official forks
 
-To run the app from source, just clone the code and run this in your terminal:
-
-```
-npm install
-npm run build
-npm start
-```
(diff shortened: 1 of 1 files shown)


astorije:
@bews, do you mind re-opening this on [The Lounge](https://github.com/thelounge/lounge) as well? We are missing this over there and that would be a great addition!

I believe this is the case, but just double-checking this:
Are the colors in the action strings (joins, quits, ...) the same than the ones for message authorship?
That is, if I join and then post a message, will the color be the same?

Also, careful, you have an unintended commit here :-)

bews:
> Also, careful, you have an unintended commit here :-)

Already noticed :)

astorije:
> Already noticed :)

Now you have 2 :-)

```sh
git rebase -i master
# Remove the last 2 commits and save
git push --force
```

dgw:
Ah, the perils of making PRs from the master branch.

bews:
Recreated from branch - #664

PR #664, opened 2016-05-10
Fix: colored nicknames for new info messages design (x2)

bews:
Now they're colored too (if colored nicknames option is enabled).


astorije:
Thanks for this, your fix inspired me to fix the same issue on The Lounge: https://github.com/thelounge/lounge/pull/317 :-)

Issue #662, opened 2016-04-28
Delayed join / command-order


samboman:
I have a list of Channels I auto join. And I have setup a commands-entry in the user configuration file to identify myself.

```
"commands": [
  "/msg NickServ identify my-pass"
],
```

The problem are channels that require registered nicks to join (eg. #openvpn on freenode). The commands seems to be sent after the join-commands.

//Sam

samboman:
I simply fixed this by using:
`"password": "my-pass"`

Maybe a better solution!

//Sam

erming:
Glad to hear you solved it!

Just remember that the `password` property is not the same as NickServ identify. And to make sure that `commands` property wasn't broken, I tried using
```
"commands": ["/msg #chan test"]
```
and that worked for me.

dgw:
I think the issue wasn't so much commands not being sent as it was commands being sent too late. The client tries to join channels before the command to identify with NickServ is sent, and thus is denied joining channels that require the use of a registered nickname.

samboman:
@erming Yes, it worked for me... The command worked:
```
"commands": ["/msg NickServ identify my-pass"]
```
But as @dgw commented, its a timing issue... Joins happens before Identify are done. Is there a /msg wait 1000ms command to delay?

Like having:
```
"commands": ["/msg NickServ identify my-pass",
  "/wait 1000",
  "/join #authenticatedChan"]
```

erming:
@dgw is right.

However, there is already a 1000ms delay in place, but maybe it's not enough. The code for running commands and joining channels can be found here: https://github.com/erming/shout/blob/master/src/client.js#L196

@samboman if you put everything in `commands` (but remove `/wait 1000`), does that work?

samboman:
@erming No, it does not work:
`"commands": ["/msg NickServ identify my-pass", "/join #authenticatedChan"]`
It seems to fire off the commands to quick... Before they are finished.

PR #665, opened 2016-05-10
Reconnection support.

bews:
Based on code from #541.

Changes: exponential reconnection attempts fixed; added reconnect on ping timeout.

Every time reconnect happens, a message is posted to the main window:

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -31,7 +31,8 @@ var events = [
 	"quit",
 	"topic",
 	"welcome",
-	"whois"
+	"whois",
+	"ping"
 ];
 var inputs = [
 	"action",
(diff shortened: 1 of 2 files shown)


bews:
Added fix for #662 - it changes already modified code, so I committed it here.

Problem was in `irc.once("pong", function() {` - PING command is firing from somewhere else too (besides `irc.write("PING " + network.host);`) so this event happens before user commands are finished.

bews:
Merged all my additions in https://github.com/bews/shout/tree/bews-master - you can grab them from there till this not merged.

MaxLeiter:
If it won't be merged, why not use TheLounge and contribute fixes for others to also use?

bews:
> If it won't be merged, why not use TheLounge and contribute fixes for others to also use?

As I said in previous PR:
> when I started working with Shout I didn't know about Lounge fork, so I'll stick with this one :)

I just upload changes I did some time ago and don't plan making anything more here.

TheLounge is basically the same atm, so I don't feel like moving to it now. Maybe later, when it will have DCC and ban lists :)

astorije:
> TheLounge is basically the same atm, so I don't feel like moving to it now.

Just closing the loop and not trolling here but I felt the need to point readers of this to the [dedicated page listing the differences](https://github.com/thelounge/lounge/wiki/Differences-between-The-Lounge-and-Shout) :smile:
There is a lot in it, even though I admit lots are minor. [Next release will be huge however](https://github.com/thelounge/lounge/compare/v1.5.0...master) and there is more to come before we release. We'll probably go through the PRs you opened to integrate what isn't there already (colors in action messages is fixed for example).

Thanks for your contributions to Shout anyway, and feel free to join the `#shout-irc` and `#thelounge` IRC channels! :-)

bews:
> There is a lot in it, even though I admit lots are minor. 

But it still doesn't have the most important feature - auto reconnection :)

> We'll probably go through the PRs you opened to integrate what isn't there already

This would be really good.

astorije:
> But it still doesn't have the most important feature - auto reconnection :)

I knooow... So soon though: https://github.com/thelounge/lounge/pull/254

PR #667, opened 2016-05-11
Added Light theme.

bews:
It also has nice hover effect for every button.

### DIFF
--- a/client/themes/light.css
+++ b/client/themes/light.css
@@ -1,0 +1,54 @@
+/*
+Light theme for Shout.
+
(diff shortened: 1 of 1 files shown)


astorije:
Hey @bews, I know it's been a while since you have opened this, but if you made or consider making this theme compatible with [The Lounge](https://github.com/thelounge/lounge), it would be a great addition to the [*Themes from the community* wiki page](https://github.com/thelounge/lounge/wiki/Themes-from-the-community).

